PR #5, opened 2021-04-29
Add documentation on installing dependencies for tools/develop-svelte

leafOfTree:
Hi, josdejong
Great job! 
After cloning this repository, it took me some time to find out that I also needed to install dependencies for tools/develop-svelte to run `npm start` successfully. So I added documentation for it. Please check and see if necessary. Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,7 @@ Install dependencies:
 
 ```
 npm install
+(cd tools/develop-svelte && npm install)
 ```
 
 Start watcher and demo project (at localhost:5000):


josdejong:
Ah, yes thanks 👍 

Actually, I expect the current setup is not really the right one, for example I'm not sure whether for Svelte I should link to the source code or the bundle. I'm waiting a little bit for sveltekit 😄  (See https://github.com/sveltejs/kit/issues/518 and https://github.com/sveltejs/kit/issues/141#issuecomment-813307214)

Issue #23, opened 2021-10-06
Issue with pressing Enter when using with React


tylerjlawson:
This is an issue in the [sandbox](https://codesandbox.io/s/svelte-jsoneditor-react-59wxz?file=/src/App.js):

When you are in Code mode and press enter it resets the cursor to the top of the editor. I am assuming this has something to do with the implementation of the value or onChange in the ace editor, but would appreciate help with debugging this.

![Kapture 2021-10-06 at 13 46 26](https://user-images.githubusercontent.com/26290074/136256351-0e7191ad-2f54-4295-826d-34ebe68bbc09.gif)

josdejong:
Thanks for reporting. I first tried in Firefox, there the issue doesn't happen. On Chrome I can reproduce your issue.

tylerjlawson:
So I found out that the issue is related to the controlling of `content` on the AceEditor. When you remove the `content` prop from JsonEditor and allow the onChange to be called on the uncontrolled component it edits flawlessly without this issue. With this in mind I believe it would be best to add a new prop to `JSONEditor` called `defaultContent` which would set the ace editor value and the tree mode value on the initial render, but would not continue to set the value as changes are made in the editor.

I'd say that uncontrolled should also be the recommendation for the React demo as well; this thread https://github.com/facebook/react/issues/955 shows a long running confusion with controlled inputs with regards to state in React, so considering the editor works as expected and can pass updates when uncontrolled, I think that makes most sense.

josdejong:
Just to have some context, I want `svelte-jsoneditor` not only to work with React, but want to have it work in the following cases:

- in Svelte, using 2-way binding (`bind:content`)
- in Svelte, React, Vue, Angular, etc using 1-way binding (`content` and `onChange`)
- in plain JavaScript using methods `get()`, `set()`, `update()` (uncontrolled component)

Having a mix of controlled and non-controlled use cases makes things complicated indeed. I'm afraid there is no way around it except when removing some of the use-cases above.

I've done a bit of debugging in the sandbox, here an adapted version with some logging that gives insight:

https://codesandbox.io/s/svelte-jsoneditor-react-issue-23-yfobq?file=/src/App.js

What I noticed is that when you press enter in code mode, in most of the cases this will trigger two changes: first a change with the new return character added, directly followed by a change that has added indentation to the new line.

In React, the `useEffect` (and render?) is executed asynchronously, which causes a nasty race condition in this case. So what happens is: 

[...]

PR #24, opened 2021-10-07
fix: Add defaultContent to JSONEditor

tylersayshi:
This will allow react or other frameworks to initialize the editor with content and still recieve
updated content through onChange, but not continue to control the value as it is changed.

Relates to #23


josdejong:
Why isn't the current `content` property sufficient for this? You decide what you pass there, and whether you want to update this property or not.

tylerjlawson:
Very good point with regards to the `value`. What I found is probably better for this situation is to edit the sandbox to accept the prop on the react side, then have it not update, leaving the editor uncontrolled.

https://codesandbox.io/s/svelte-jsoneditor-react-forked-td9to

PR #33, opened 2021-11-10
dependency debug: cjs -> esm (fix vite)

### LINKED ISSUE #31: vite: TypeError: Cannot set properties of undefined (setting 'j')
TLDR: how to actually use this library? 0__o
... project template?

when i run the example [json_schema_validation.svelte](https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/examples/json_schema_validation.svelte), i get the error

> Uncaught SyntaxError: The requested module '/node_modules/.pnpm/debug@4.3.2/node_modules/debug/src/browser.js?v=ab0664fe' does not provide an export named 'default'

from

https://github.com/josdejong/svelte-jsoneditor/blob/8e5aecf4b0ac29c97d545098f037d7f027182558/src/lib/components/modals/popup/AbsolutePopup.svelte#L4

### backtrace

```js
// node_modules/debug/src/browser.js

module.exports = require('./common')(exports);
```

```js
// node_modules/debug/src/common.js

function setup(env) {
        createDebug.debug = createDebug;
        // ...
        return createDebug;
}

module.exports = setup;
```

looks like a problem with vite and commonjs ...

### reproduce

```sh
npm i -g pnpm
pnpx create-vite frontend --template svelte
cd frontend
pnpm install svelte-jsoneditor
cd src
curl -L -o App.svelte https://github.com/josdejong/svelte-jsoneditor/raw/main/src/routes/examples/json_schema_validation.svelte
sed -i 's/\$lib/svelte-jsoneditor/' App.svelte
cd ..
npm run dev
```

### related

also pnpm does not run the `prepare` script when i install from git ...

```
npm rm svelte-jsoneditor
pnpm i https://github.com/josdejong/svelte-jsoneditor.git

stat node_modules/svelte-jsoneditor/dist
stat: cannot statx 'node_modules/svelte-jsoneditor/dist': No such file or directory
```

milahu:
fix #31 and make svelte-jsoneditor work with vite

error was

> TypeError: Cannot set properties of undefined (setting 'j')

and

> Uncaught ReferenceError: process is not defined

see https://github.com/visionmedia/debug/pull/852#issuecomment-965475376

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,13 +26,14 @@
     "prepare": "husky install && cd tools && node generateEmbeddedAceJsonWorker.js && cd ..",
     "release": "npm run lint && npm test && standard-version && npm run package && git push && git push --tag && npm publish package/"
   },
+  "__comment_dep_debug": "workaround debug-esm from https://github.com/visionmedia/debug/pull/852",
   "dependencies": {
     "@fortawesome/free-regular-svg-icons": "^5.15.4",
     "@fortawesome/free-solid-svg-icons": "^5.15.4",
     "ace-builds": "^1.4.13",
     "ajv": "^8.6.3",
     "classnames": "^2.3.1",
-    "debug": "^4.3.2",
+    "debug": "github:milahu/debug#28e27c8b4249fec2f778cf61f13e18e8c814822a",
     "diff-sequences": "^27.0.6",
     "immutable-json-patch": "^1.1.2",
     "json-source-map": "^0.6.1",


josdejong:
Thanks. Good to know that a fixed version of debug actually solves the issues you have. 

I'm a bit surprised that only `debug` gives an issue in your case, I identified more dependencies that give an issue: https://github.com/josdejong/svelte-jsoneditor/blob/main/src/lib/config.js

I'm hesitant to include a git branch as dependency: if you happen to remove this branch from your git repo, the library would be broken. 

Are there alternative libraries that we can look into to replace debug?

milahu:
> I'm hesitant to include a git branch as dependency: if you happen to remove this branch from your git repo, the library would be broken.

that is a valid concern ...
but also npm is not safe, just remember [the left-pad scandal](https://www.theregister.com/2016/03/23/npm_left_pad_chaos/)

options:

1. you publish my branch (actually @Konakona-chan's branch) as npm package
2. you use the unmaintained (but [mostly](https://github.com/ds300/patch-package/pull/302) working) [patch-package](https://github.com/ds300/patch-package) to patch the npm version of debug

> Are there alternative libraries that we can look into to replace debug?

should be

let me [google that](https://www.google.com/search?q=npm+debug+logger+library) for you ...

* https://www.npmjs.com/package/debug-logger
* https://www.npmjs.com/package/better-logging
* https://www.npmjs.com/package/js-logger
* https://www.npmjs.com/package/log
* https://blog.logrocket.com/comparing-node-js-logging-tools/
  * https://www.npmjs.com/package/winston
  * https://www.npmjs.com/package/pino
  * https://www.npmjs.com/package/bunyan
  * https://www.npmjs.com/package/morgan
  * https://www.npmjs.com/package/npmlog
* probably 100 more

josdejong:
Thanks. I did some googling too, but I don't really see a lightweight solution which supports colored namespaces (I like that a lot). 

I simply created a solution myself, see #34 . It's really no rocket science. What do you think?

milahu:
> It's really no rocket science. What do you think?

simple but effective : )

josdejong:
Will close this issue now in favor of the merged and published #34. Thanks anyway!

PR #34, opened 2021-11-27
fix: replace `debug` with a util debug function

josdejong:
@milahu can you have a look at this PR?

For reference: #33, #31.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,6 @@
         "ace-builds": "^1.4.13",
         "ajv": "^8.8.2",
         "classnames": "^2.3.1",
-        "debug": "^4.3.2",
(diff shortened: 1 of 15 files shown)


josdejong:
I would like to see if it is possible to use some environment variable of Svelte to detect whether running in development or production, and use that to enable/disable logging. It would be cool to see if we can utilize dead code elimination of all the debug lines when in production.

milahu:
lgtm : )

> dead code elimination

can we micro optimize this later? ; )

josdejong:
Thanks.

> can we micro optimize this later? ; )

😂  yes. This new solution already has much less lines of code, so it's already an improvement anyway.

josdejong:
Published now in `v0.3.6`

Issue #38, opened 2021-12-13
Babel, svelte-jsoneditor >0.3.5 error: "You may need an additional loader to handle the result of these loaders."


dranitski:
For some reason `svelte-jsoneditor@>0.3.5` crashes in react starter app just after being imported:

Steps to reproduce:
- `npx create-react-app my-app`
- `cd my-app`
- `yarn add svelte-jsoneditor@0.3.5`
- add this line to `src/App.js`: `import {JSONEditor} from 'svelte-jsoneditor/dist/jsoneditor.js';`
- run npm start
- see the error:
```
Failed to compile
./node_modules/svelte-jsoneditor/dist/jsoneditor.js 13423:12
Module parse failed: Unexpected token (13423:12)
File was processed with these loaders:
 * ./node_modules/react-scripts/node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
|     label: "descending"
|   }];
>   let f = u?.filter?.path ? I(u.filter.path) : null,
|       p = u?.filter?.relation ? h.find(e => e.value === u.filter.relation) : null,
|       m = u?.filter?.value || "",
This error occurred during the build time and cannot be dismissed.
```

I've noticed downgrading to svelte-jsoneditor@0.3.4 seems fixing the issue for now.
Thanks in advance!

josdejong:
Thanks for reporting, and for the clear reproduction steps. 

I think that optional chaining `?.` is giving an issue here, I started using that in the TransformWizard in v0.3.5. So far the bundle isn't transpiled and I'm happily using all available JavaScript features 😁. 

I suppose if we want to prevent this issue we'll have to configure babel and transpile the bundle to be compatible with old browsers. Help would be welcome!

dranitski:
Seems like we need this:

> If you’re using Babel, @babel/plugin-proposal-optional-chaining, @babel/plugin-proposal-nullish-coalescing-operator are available.

I will try it in my current project and let you know if it will solve the issue.

josdejong:
Yes indeed. And `svelte-jsoneditor` currently doesn't have babel transpilation in place but we could add that to the rollup config. That would solve it for all consumers of the library.

josdejong:
I just tried out with the latest `create-react-app@5.0.0` (which comes with Webpack 5) which was released 8 days ago on Dec 14th, that also solves the issue.

brianpos:
Is there an equivalent Vuejs (or nuxt version)?
[...]

Issue #19, opened 2021-09-02
Make it TypeScript compatible


apirogov:
I am using typescript in my project and e.g. when I use "createAjvValidator" and pass it a schema, TypeScript complains because due to the comment in the code it thinks there is some JSON type and my data is not of that type.

It would be great if the exposed functions would be made properly TypeScript compatible, or at least fixed in a way that TypeScript is not confused and treats everything as `any` instead of non-existing types.

josdejong:
Yes, I would like to convert the source code to TypeScript. 

This requires some fiddling with the build setup to get this all working though. Help would be welcome.

cloakedninjas:
So - now that v6 is released and in lieu of source being in TypeScript - will there be .d.ts file provided? [DefinitelyTyped](https://www.typescriptlang.org/dt/search?search=svelte-jsoneditor) is missing this project

josdejong:
Yes the plan is to make this a TypeScript project. The Svelte setup is TypeScript-ready but there where some issues getting the mocha tests working with TypeScript files.

Help would be welcome.

josdejong:
I've fixed the setup so it all works with TypeScript. I've converted a couple of files already, though most files still have to be converted. We'll get there 😄

cloakedninjas:
Thanks for the update, but ideally the `types.d.ts` file gets placed alongside the dist file

Doing
```
import { JSONEditor } from 'svelte-jsoneditor/dist/jsoneditor';
```

Still results in `TS7016: Could not find a declaration file for module 'svelte-jsoneditor/dist/jsoneditor'`

And once that's created, a module or interface will be needed for the whole library too, e.g.

```
export declare interface SvelteJsonEditorInitOpts {
    target: HTMLElement;
    props?: JsonEditorOptions;
}
```

Cheers 👍

josdejong:
Yes indeed, there is work to be done. A couple of things:

- Step one was to make the setup (build tools and testing) work with TS, that is now down, 
- So now we can gradually convert all code to TS (working on that)
- The Svelte output has proper TS definitions and should work as is
- The vanilla JS bundle output `dist/jsoneditor.js` is missing TS definitions. 
  - I want to split the npm package in two (one for Svelte, one for Vanilla JS), in order to prevent conflicts between the two and issues with build tools because of that
  - I need to get proper type definitions out of the rollup bundler that generates `dist/jsoneditor.js`. Fiddling around with that.

josdejong:
[...]

PR #85, opened 2022-05-26
Fixes conditional chaining operator issues for webpack 4 users

gerardosabetta:
Hey there this fixes issue #38.

I need this for my project, hope it also helps folks out there.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6232 +1,19 @@
 {
   "name": "svelte-jsoneditor",
   "version": "0.3.54",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
   "requires": true,
-  "packages": {
-    "": {
-      "name": "svelte-jsoneditor",
-      "version": "0.3.54",
-      "license": "ISC",
-      "dependencies": {
-        "@codemirror/basic-setup": "^0.19.3",
-        "@codemirror/commands": "^0.19.8",
-        "@codemirror/highlight": "^0.19.8",
-        "@codemirror/history": "^0.19.2",
-        "@codemirror/lang-json": "^0.19.2",
-        "@codemirror/language": "^0.19.10",
-        "@codemirror/lint": "^0.19.6",
-        "@codemirror/search": "^0.19.10",
-        "@codemirror/state": "^0.19.9",
-        "@codemirror/view": "^0.19.48",
-        "@fontsource/fira-mono": "^4.5.8",
-        "@fortawesome/free-regular-svg-icons": "^6.1.1",
-        "@fortawesome/free-solid-svg-icons": "^6.1.1",
-        "ajv-dist": "^8.11.0",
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Gerardo, looks good.

Three small remarks:

1. Is there a specific reason to use `getBabelOutputPlugin` and not `babel`? (I hadn't seen it before. Does the source map still work with that transpilation step?)
2. I just published a new version of `svelte-jsoneditor`, can you fix the merge conflicts in this PR?
3. Can you use exact version numbers in the package.json devDependencies?

gerardosabetta:
Hey Jos, yes I will resolve those conflicts. I used `getBabelOutputPlugin` because I've read the following in the docs:
```
You can run @rollup/plugin-babel on the output files instead of the input files by using getBabelOutputPlugin(...)
```

So I might be wrong but I think this makes babel work on the bundled file instead of the src files. I can change it to `babel` if you want I think it would be less overhead

josdejong:
> So I might be wrong but I think this makes babel work on the bundled file instead of the src files. I can change it to babel if you want I think it would be less overhead

Right now the rollup config is only used to generate this single bundled file. The original src files should indeed not be touched, but would rollup really do that otherwise? I would have to do some digging to on the behavior of rollup, I'm not an expert 😅 . Most important is that the outputted bundle works, is transpiled, and that the source map works.

gerardosabetta:
To be honest I haven't tested the source maps I will set a breakpoint and see if they work. I will also compare using `babel` and `getBabelOutputPlugin` I'm also not an expert when it comes to rollup 😅

gerardosabetta:
Hey @josdejong I fixed the conflicts. I'm still not 100% sure that my change does not affect the sourcemaps and I'm having trobule testing those can you or someone please help me assert that the sourcemaps are functional? Thanks in advance.

I've also switched to exact version numbers in the `package.json`

josdejong:
Thanks for the updates.

I tried out with `babel` instead but can't get that working. So let's just go with `getBabelOutputPlugin` 😁 .

About the source maps: I just noticed that these are broken since setting up typescript (see https://github.com/josdejong/svelte-jsoneditor/issues/19#issuecomment-1141280161), so there is work to be done on that anyway, and this PR is not introducing a regression in that regard.

Will merge your PR now

gerardosabetta:
Oh I forgot to mention that I've also tried `babel` and the generated code still had the conditional chaining operator that's why I kept `getBabelOutputPlugin` instead.

With regards to the sourcemaps thanks for confirming that they weren't working in the first place, I wanted to make sure it wasn't my change that break them. 

Thanks for merging the PR Jos! 🥳

josdejong:
😂 yeah well I bumped into the same issue I guess.

I've now published your fix in `svelte-jsoneditor@0.3.56` (I've disabled the sourcemap for the time being since it's broken anyway). Thanks again!

PR #130, opened 2022-08-26
improve readOnly option

fonov:
I notified that keyboard  is appeared on mobile device even `readOnly` is set `true` when user tap on editor. 

Set `EditorView.editable -> false` disable appeared keyboard on tap or select

### DIFF
--- a/src/lib/components/modes/textmode/TextMode.svelte
+++ b/src/lib/components/modes/textmode/TextMode.svelte
@@ -93,6 +93,7 @@
   const readOnlyCompartment = new Compartment()
   const indentUnitCompartment = new Compartment()
   const tabSizeCompartment = new Compartment()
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks Fonov for the fix!

I think the function `updateReadOnly` should also be adapted.

Thinking about this: probably we should remove the `EditorState.readOnly` and instead just use `EditorView.editable`?

josdejong:
@fonov what do you think?

josdejong:
I've tried out only using `EditorView.editable`, but then certain actions are still working, like Ctrl+V to paste something. So now I've defined both `EditorView.editable` and `EditorState.readOnly`.

josdejong:
fixed now in v0.7.4, thanks again!

Issue #98, opened 2022-06-14
navigator.clipboard is undefined on non-secured orgin (http)


hfutyezi:
i found this problem when my website is not https ,and some browser will have this error.
navigator.clipboard  return undefined

hfutyezi:
My action is to copy text

josdejong:
Ah, that's interesting. For security reasons, `navigator.clipboard` is not available when you're on a non-secure connection.

I see a workaround here that may work: https://stackoverflow.com/a/33928558/1262753

josdejong:
The fix by @cloydlau has been published in `v0.7.5`, thanks again.

PR #144, opened 2022-09-16
fix: #98 , and make copy more compatible

cloydlau:


josdejong:
Thanks for workgin out this fix Cloyd 👍. It looks good at first sight. I'll do some thorough testing tomorrow.

josdejong:
I've done some testing @cloydlau, and had a good look at the code. I like that the code is a self-contained function, very neat!

Two small feedbacks:

1. Can you add a type declaration `copyToClipBoard (text: string)` ? (I'll see if I can configure TypeScript more strictly so it will yell about this itself)
2. Can you run `npm run format` to fix some linting issues? (I'll configure a Github Action to run `npm run lint` so any issues will popup automatically)

And one question: do you know whether pasting works on non-secure http origins? If not, is there a similar solution possible you think?

cloydlau:
Sure, all done. And for HTTP, I've tested it before in this way:

image

josdejong:
Thanks for the updates, looks good!

I just came across one more issue though I'm not sure if we should fix it: after copying using the "legacy" solution, the editor loses focus. What do you think, should we restore focus afterwards?

cloydlau:
Indeed! The "legacy" solution is for outdated browsers and unsecured websites, in which case we may only gurantee the core functions.  So my personal view is to leave it be.

josdejong:
Yes that makes sense. Thanks!

josdejong:
The fix has been published in `v0.7.5`

cloydlau:
My pleasure. Actually I made this PR because someone raised [a same issue](https://github.com/cloydlau/json-editor-vue/issues/8) in my repo [json-editor-vue](https://github.com/cloydlau/json-editor-vue) (an universal Vue wrapper for `svelte-jsoneditor`). Not the first time lol. If I might venture to ask that would the doc of `svelte-jsoneditor` take section like 'Related projects'? The doc indeed provides Vue example yet no two-way binding, either no example for Vue 2. So this 'Related projects' might be helpful for the Vue users of `svelte-jsoneditor`. It's my pleasure to make a PR.

josdejong:
Ah, nice, thanks! 

I've added a section linking to wrapper libraries in the readme, right now only containing you Vue library: https://github.com/josdejong/svelte-jsoneditor#wrapper-libraries-for-specific-frameworks

cloydlau:
Thanks! It really encouraged me.

josdejong:
You're welcome! Thanks for making the library easier to use in Vue 😊 . 

Have a nice weekend.

Issue #82, opened 2022-05-19
can we customise the menu options here, if yes, can i know how can we do it.


Josnaaa:
As we already have a insert, remove, transfer. Are we allowed to add or delete this options???

josdejong:
Right now there is an option to customize the main menu with `onRenderMenu`. There is no option yet to customize the context menu. Is that what you mean?

Josnaaa:
the context menu is nothing but the below right?? If yes , then yes that's what i meant.
![Screenshot from 2022-05-19 19-22-14](https://user-images.githubusercontent.com/95766090/169310487-19066bfc-4fd3-4e03-9d16-cb1c4da0eeee.png)

Josnaaa:
okay, Thank you for the information.

1pone:
I'm really looking forward to this feature, in fact I'm currently using the react version of the editor, but in order to implement the custom menu I had to folk a copy of the source code to try and modify the svelte code myself, however that's not something I'm good at and it's very annoying, I'll be relieved the day the feature is implemented!

josdejong:
👍  help implementing `onRenderContextMenu` would be welcome.

josdejong:
A new option `onRenderContextMenu` is now available in `v0.21.0`.

toolsh:
> A new option `onRenderContextMenu` is now available in `v0.21.0`.

good to know that, is there any example / demo code for usage of `onRenderContextMenu`

josdejong:
There is an example here: [/examples/custom_context_menu_buttons/+page.svelte](https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/examples/custom_context_menu_buttons/%2Bpage.svelte).

At this moment there is still is an issue with the new `onRenderContextMenu`, see #379. It's fixed but not yet published.

PR #159, opened 2022-10-08
New feature added: custom drop down button with custom items possible

mercurio123:
Regarding issue #82: "Implement an onRenderContextMenu to customize the context menu #82" I modified the context menu in such a way that it is possible to show a custom dropdown button. The items of this special button can be customized. If one does not want this functionality, one can just leave the button invisible. The code is backward compatible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,7 +76,7 @@
         "ts-node": "10.9.1",
         "tslib": "2.4.0",
         "typescript": "4.8.2",
-        "vite": "3.1.0-beta.1"
+        "vite": "^3.1.0-beta.1"
       }
     },
     "node_modules/@ampproject/remapping": {
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "svelte-jsoneditor",
+  "name": "@mercurio123/svelte-jsoneditor",
   "description": "A web-based tool to view, edit, format, transform, and validate JSON",
   "version": "0.7.1",
   "type": "module",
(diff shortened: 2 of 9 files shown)


josdejong:
Thanks for picking this up @mercurio123 👍 

I think this is a step into the right direction, but I would like to go a step further. The solution in this PR allows adding a custom dropdown button, but it does not allow hiding or rearranging buttons for example and cannot be used in React since you have to provide a Svelte component. 

What I have in mind is a solution similar to the `onRenderMenu` function. This callback gets a list of items (as data, not as Svelte components), and can output a changed list with items. That has some advantages: you can add/remove/move items as you like, and insert new items at the start or end or somewhere halfway, and it can be used in any framework, not only Svelte.

To make a more generic solution similar to `onRenderMenu` working we'll need to refactor ContextMenu.svelte to work based on a list with data items, and after that we can implement a callback that allows to change the list before rendering. Does an approach like that make sense to you?

mercurio123:
Thank you for your feedback @josdejong

That's correct. At first, I wanted to do it as it has been done with the menu bar. Since I'm new to svelte and JavaScript I constrained it to implement a customizable dropdown button with custom snippets items.

Yes, it makes sense, I will change ContextMenu.svelte as it has been done for the onRenderMenu so that a similar function for the context menu can be implemented.

josdejong:
👍 

If you're new to both JavaScript and Svelte, I think this will be a relatively complex undertaking. Feel free to give it a try, but if it is too complicated just let me know (It's on my list but I'm not sure when I'll have time for it).

mercurio123:
@josdejong I will, but I will come to it after I have finished the other parts of my project.

josdejong:
Thanks for the update. No problem at all. I'll close this PR for the time being, and maybe it's easiest to create a new PR when you are working on it again.

lperry25:
@josdejong @mercurio123 was this feature ever picked up into a new PR?

Issue #160, opened 2022-10-10
Convert Structure throws an error, can't convert to Object or Array


wernerstucky:
On the classic JSON editor one of the functions we used a lot was converting between types.
It seems this conversion is not working on the new version (see video clip)

If you try to convert a string type into an Array or Object you get a the following error : 
```
SyntaxError : JSON value expected but reached end of input at position 0

```

[jsoneditor convert type bug-2022-10-10_07.38.45.webm](https://user-images.githubusercontent.com/2656247/194804465-5ec24a6a-b3e2-4695-8c89-bc7d5daccc0c.webm)

josdejong:
Fixed now in `v0.7.11`

PR #161, opened 2022-10-10
Update jsonUtils.ts

wernerstucky:
Proposed fix for Issue : #160 converting a non-JSON string to Object or Array. This brings back functionality that was present in the classic JSONEditor We catch JSON parse errors and return a reasonable Object / Array.  This is an improvement on the classic editor which used to purge the string value.

### DIFF
--- a/src/lib/utils/jsonUtils.test.ts
+++ b/src/lib/utils/jsonUtils.test.ts
@@ -1,5 +1,5 @@
 import type { JSONParser } from '$lib/types.js'
-import { deepStrictEqual, strictEqual, throws } from 'assert'
+import { deepStrictEqual, strictEqual, throws, equal, deepEqual } from 'assert'
 import { parse, stringify } from 'lossless-json'
 import {
   calculatePosition,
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks @wernerstucky , I like your PR and inserting the original value as array item or object value 👌.

I have a few feedbacks:

1. Naming the new key `"0"` sounds a bit odd to me. How about naming it `"value"` or `"key"` or `"New prop"` or something like that?
2. Can you please write unit tests to validate the new behavior?
3. Does this new behavior also work for numbers, booleans and null? If not, can you make that work too?

wernerstucky:
Thanks @josdejong  I'll like those ideas. Let me work on that and we take it from there.

wernerstucky:
@josdejong over to you, 

1. value gets added under "value"
2. added a few tests for various values
3. added default return to both array and object cases which will handle boolean, number, etc

josdejong:
Thanks for the updates, this looks good 👌 

Can you please run `npm run format` to fix the linting issues? (you can probably configure your IDE to apply formatting when saving a file).

wernerstucky:
Looks like those linting issues are fixed.  I've installed the Prettier extension thanks for the tip!

josdejong:
💪 thanks!

josdejong:
Published now in `v0.7.11`

PR #164, opened 2022-10-13
Implement a path editor in the Navigation Bar

### LINKED ISSUE #13: Allow copy node's path
> Could be copy on ContextMenu or emit on `mousedown`.

josdejong:
Fixes #13 

- Implements a path editor where you can see and modify the path, copy, paste, and just type what you want.
- A special copy button on the right side to conveniently copy the path to clipboard
- Throws and error when the path is invalid or does not not exist in the document
- Customizable format: defaults to JSON Path, but you can configure to use for example JSON Pointer instead using the new `pathParser` option.
- Export two new util functions: `parseJSONPath` and `stringifyJSONPath`

afbeelding

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,7 +223,9 @@ const editor = new JSONEditor({
   </div>
   ```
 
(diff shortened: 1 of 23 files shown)


josdejong:
Published now in `v0.7.10`

PR #178, opened 2022-10-30
fix: showing `cursor: grab` in read-only mode

cloydlau:
Hello,
Since sorting is not allowed in read-only mode, there should be no `cursor: grab` style either.
Feel free to correct me if you think it's not the best way to solve this 😊.

### DIFF
--- a/.github/workflows/ci-test.yaml
+++ b/.github/workflows/ci-test.yaml
@@ -20,9 +20,9 @@ jobs:
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
@@ -34,10 +34,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v1
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
-          node-version: 16.x
+          node-version: 18.x
       - run: npm ci
       - run: npm run lint
         env:


josdejong:
Thanks @cloydlau , good point about the grab style cursor!

The solution that you propose here is to disable selecting contents in read-only mode. I do think think though that it is fine and valuable to still be able to select contents in a read-only document and then copy them. So instead, I think we should add a class like `class:jse-readonly={context.readOnly}` and disable `cursor: grab` style in CSS when that class is present. What do you think?

cloydlau:
That's better. I missed the copy thing.

Now I have two solutions:
1. Using class + scss variables
2. Using css variables: Lesser code but not supported in Chrome@48 or earlier

Which one do you prefer?

josdejong:
Why would there be need for a scss variable or css variable? I'm not sure but isn't this a regular CSS condition along the lines of 
`&:not(.jse-readonly) { cursor: grab }`?

josdejong:
About your solution in #181: now I understand what you mean. I haven't see this kind of approach before, using a CSS variable and setting that via `style`. I'm not sure if I like this approach, though maybe I have to get familiar with it. I doubt whether CSS variables are meant to be injected in all kind of places via `style`. In general I try to avoid using `style` and use class names instead.

> Hmm... but I'm not sure whether `&:not(.jse-readonly) { cursor: grab }` can handle deeply nested SCSS code.

Yes that is tricky, it would mean more or less duplicating all the CSS in the `@mixin json-node-selected-mixin`, wrapped in this `.jse-readonly` class. Using `!important` we can keep things quite simple, see #183. It works but it's not very pretty.

Now, this difficulty comes from the `@mixin json-node-selected-mixin` mixin being way to complicated right now: it uses `:global` and `!important`, too many occurrences of `background: $selection-background`, and deep nested CSS rules which require knowledge of child components, which is ugly. I think the underlying issue is that I basically add a class name `jse-selected` at the root of the selection only, and then need complex CSS rules to highlight the right nested parts. I would like to try out a different approach: attach a `jse-selected` class to all nested parts instead that need highlighting. I would like to address this refactoring separately from the issue with `cursor: grab` though, and take a pragmatic approach to solve the grab-issue right now.

cloydlau:
Using CSS variables in `style` tag is not unusual. See what ElementPlus do in which `el-switch` component: https://github.com/element-plus/element-plus/blob/dev/docs/examples/switch/basic.vue. But surly it's up to you, that's all right. 😊
The other 'class + SCSS variables' solution I mentioned before has lesser code than https://github.com/josdejong/svelte-jsoneditor/pull/183, and what's more it's less duplicating. I think it's worth a try and I will be willing to make a new PR for it. But if you think https://github.com/josdejong/svelte-jsoneditor/pull/183 will do the job that's ok to me too.

josdejong:
Yes, it is an interesting approach.  I learn something new every day :). Maybe I simply have to get used to it. I'm not a CSS expert though, and not all shiny new approaches turn out to be a good idea (having the sentence "Frequently inserting CSS rules forces the browser to do a lot of extra work" from the following recent article in the back of my mind 🤔: https://dev.to/srmagura/why-were-breaking-up-wiht-css-in-js-4g9b). So in case of doubt I simply try to keep my CSS simple, old-fashioned, and boring 😁.

So... for now I say let's go for #183, and in a second step refactor the whole `@mixin json-node-selected-mixin`. When doing that I'll keep your techniques in mind (css variables or scss variables), but since things will go upside down then I'm not sure at this point solution we need in the end. Thanks for all the thinking along!

cloydlau:
Sure. The performance should be ok as long as it does not cause a Reflow though. I'll say you are quite progressive considering you are using `vite`. A refactor will be helpful 👍.

josdejong:
👍

PR #186, opened 2022-11-07
fix: unexpected scrollbar and gap in text mode

cloydlau:
As shown in the figure:

### DIFF
--- a/src/lib/components/modes/textmode/TextMode.scss
+++ b/src/lib/components/modes/textmode/TextMode.scss
@@ -32,13 +32,13 @@
 
     :global(.cm-editor) {
       flex: 1;
-      overflow: hidden;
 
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks for your PR. 

I'm a bit confused, when I open the example you show in your screenshot, I do not see this problem (I do not see a scrollbar nor white space at the bottom). And when opening this PR, it introduces a new problem: there is no scrollbar when it should have one, for example when you change the height of the editor to 200px, or paste al larger JSON document.

cloydlau:
Investigated this and it turned out that a Chrome Plugin did this. Sorry, it's my bad 😓. I should've tested it in different browsers.

josdejong:
😂  yes can happen, no problem. Good to hear you've found the cause.

Issue #53, opened 2022-04-04
How to import bundle from plain JS


Mark-81:
I want to use this library in a small PHP/JS project.
Reading the docs I'm trying to follow the examples for plain javascript.

This line:

    import { JSONEditor } from 'svelte-jsoneditor'

does not work because the browser (Firefox) requires the module name starts with `/`, `./` or `../`.
So I changed it to an absolute path:

    import { JSONEditor } from '/node_modules/svelte-jsoneditor/dist/jsoneditor.js'

Now I have a 404 (even the path is correct - PHP Storm can retrieve the file) and another error:

> Loading module from “http://mydevice.local/node_modules/svelte-jsoneditor/dist/jsoneditor.js” was blocked because of a disallowed MIME type (“text/html”).

I copied and paste [this](https://github.com/josdejong/svelte-jsoneditor/blob/main/examples/browser/basic_usage.html) example in my HTML page. I only adjusted the path of the module to match the actual position on my machine.

Again, if I copy to the bundle file to `/js/jsoneditor.js` and change the include to a common script tag like this:

    <script type="text/javascript" src="/js/jsoneditor.js">

I get this error:

> Uncaught SyntaxError: export declarations may only appear at top level of a module [jsoneditor.js:55:43779](http://mydevice.local/js/jsoneditor.js)

josdejong:
I suppose you have to make sure the dist/jsoneditor.js file is hosted correctly by your PHP backend, and can be fetched by the frontend. So first make sure your version of the url `/node_modules/svelte-jsoneditor/dist/jsoneditor.js` actually works, you can try if you can load this in your browser and see if it returns the JavaScript bundle. Alternatively, you can load the bundle from a CDN if that is acceptable for your project.

Here is a working example that fetches the bundle from a CDN: https://jsbin.com/zotiler/edit?html,output

josdejong:
> Again, if I copy to the bundle file to `/js/jsoneditor.js` and change the include to a common script tag like this:
> 
> ```
> <script type="text/javascript" src="/js/jsoneditor.js">
> ```
> 
> I get this error:
> 
> > Uncaught SyntaxError: export declarations may only appear at top level of a module [jsoneditor.js:55:43779](http://mydevice.local/js/jsoneditor.js)

You cannot load the library like `<script type="text/javascript" src="/js/jsoneditor.js">`, it is an ES module, see demo linked above.

Issue #165, opened 2022-10-13
vanilla-jsoneditor in commonjs


hacknlove:
I would like to be able to import vanilla-jsoneditor in my commonjs project.

In the development environment I have been able to workaround this with `import()` but somehow after building it fails.

I have tried to fork the repo and compile it myself to commonjs, but I failed.

I have also tried many other workarounds, and nothing has worked so far.

My two last options are:
1. ask this.
2. look for any alternative library

But I like svelte-jsoneditor, you nailed it. It's real good UX.

So, please, could you bundle it in commonjs too?

Thanks.

josdejong:
Thanks for your suggestion, glad to hear you like the library :)

I'm open to publishing for example an UMD (or CommonJS) version of the library if there is enough need for it. I'm a bit hesitant though: I haven't done that so far, hoping we don't need it nowadays. I would love to get the JavaScript world out of the difficult CommonJS/ESM hybrid situation where we currently are. I think we should all move to ESM as quick as possible.

For the time being, here some ideas for workarounds:

1. The `index.js` file in `vanilla-jsoneditor` is a single, standalone file with a single `export` at the end of the file. To turn it into CommonJS, all you have to do is replace the `export{...}` with `module.exports={...}`. You can write a script reading the file, use a regex to replace the export, and write it in a different file.
2. With for example browserify, you can create an UMD bundle:

    ```
    npm install --save-dev browserify esmify browser-resolve
    npx browserify ./node-modules/vanilla-jsoneditor/index.js -p esmify -o ./jsoneditor.umd.js --standalone jsoneditor
    ```

    Then you can load the output in a browser (or load as CommonJS):

    ```html
    <script src="./jsoneditor.umd.js"></script>
    <script>
      const { JSONEditor } = window.jsoneditor
      // ...
    </script>
    ```
3. There are quite some tools around to convert ESM to CommonJs, maybe there is something else useful around?

hacknlove:
Thanks, it worked.

I didn't see the export at the end of the index. It's minimized. I tried bundling it without the minimizer, but I may have not done it right because I have still got non-readable code.

But I used browserify as you said in 2.

Then I needed to add the types declaration from vanilla-jsoneditor/index.d.ts and lastly to prevent it from being used in the ssr.

:)

josdejong:
👍 that is good to hear.

[...]

PR #195, opened 2022-11-22
build: offer UMD format output for 'vanilla-jsoneditor'

cloydlau:
Hi there. 👋

At present 'vanilla-jsoneditor' only provides ES format output which cannot handle situations like:
1. CDN usage in not-that-new browsers (Chrome ≤61), see https://caniuse.com/es6-module.
2. CommonJS environment.

CDN usage support is meaningful, see: 
- https://github.com/cloydlau/json-editor-vue/issues/5
- https://github.com/josdejong/svelte-jsoneditor/issues/53

If one dependency is ES format, it compels other dependencies to use the same way (native ES modules), which makes it quite messy, see https://github.com/cloydlau/json-editor-vue#cdn-2.

I made a sheet about what output formats are provided and what do they name it by popular libs:

| lib                            | es                           | umd                           | cjs                     | iife                 |
| ------------------------------ | ---------------------------- | ----------------------------- | ----------------------- | -------------------- |
| libs using Vite Default Config | `xxx.mjs`                    | `xxx.umd.js`                  | -                       | -                    |
| Svelte                         | `index.mjs`                  | -                             | `index.js`              | -                    |
| Vue                            | `vue.runtime.esm-bundler.js` | -                             | `vue.cjs.prod.js`       | `vue.global.prod.js` |
| React                          | -                            | `umd/react.production.min.js` | `index.js`              | -                    |
| Rollup                         | `es/rollup.js`               | -                             | `rollup.js`             | -                    |
| Axios                          | `esm/axios.min.js`           | `axios.min.js`                | `node/axios.cjs`        | -                    |
| ECharts                        | `echarts.esm.min.js`         | `echarts.min.js`              | `echarts.common.min.js` | -                    |
| Day.js                         | `esm/index.js`               | `dayjs.min.js`                | -                       | -                    |
| CodeMirror                     | `index.js`                   | -                             | `index.cjs`             | -                    |

We can see that no one is left CommonJS behind, which is provided by .cjs itself, UMD or both.

What do you think? If you agree with this:
1. What file name of UMD output should be? Now I tentatively name as `index.umd.js`.
2. Since we have two outputs now, should we change the file name of ES output too?

### DIFF
--- a/rollup.config.bundle.js
+++ b/rollup.config.bundle.js
@@ -9,44 +9,64 @@ import terser from '@rollup/plugin-terser'
 import sveltePreprocess from 'svelte-preprocess'
 
 const production = !process.env.ROLLUP_WATCH
-const packageFolder = 'package-vanilla'
-const file = path.join(packageFolder, 'index.js')
+
+export const packageFolder = 'package-vanilla'
+export const name = 'vanilla-jsoneditor'
+export const moduleFile = 'index.js'
+export const umdFile = 'index.umd.js'
+
+const plugins = [
+  svelte({
+    compilerOptions: {
+      // enable run-time checks when not in production
+      dev: !production
+    },
+
+    // we want to embed the CSS in the generated JS bundle
+    emitCss: false,
+
+    preprocess: sveltePreprocess()
+  }),
+
+  resolve({
+    browser: true
+  }),
+  commonjs(),
+  json(),
+
+  typescript({ sourceMap: true, inlineSources: true })
+]
 
 export default {
   input: 'src/lib/index-vanilla.ts',
   output: [
     {
-      file,
+      file: path.join(packageFolder, moduleFile),
       format: 'es',
       sourcemap: true,
-      inlineDynamicImports: true
+      inlineDynamicImports: true,
+      // output-specific plugins can only modify code after the main analysis of Rollup has completed
+      plugins: [
+        getBabelOutputPlugin({
+          presets: ['@babel/preset-env'],
+        }),
+        // minify
+        production && terser()
+      ]
+    },
+    {
+      name,
+      file: path.join(packageFolder, umdFile),
+      format: 'umd',
+      sourcemap: true,
+      inlineDynamicImports: true,
+      globals: {
+        [name]: 'JSONEditor',
+      },
+      plugins: [
+        production && terser()
+      ]
     }
   ],
-  plugins: [
-    svelte({
-      compilerOptions: {
-        // enable run-time checks when not in production
-        dev: !production
-      },
-
-      // we want to embed the CSS in the generated JS bundle
-      emitCss: false,
-
-      preprocess: sveltePreprocess()
-    }),
-
-    resolve({
-      browser: true
-    }),
-    commonjs(),
-    json(),
-
-    typescript({ sourceMap: true, inlineSources: true }),
-    getBabelOutputPlugin({
-      presets: ['@babel/preset-env']
-    }),
-
-    // minify
-    production && terser()
-  ]
+  plugins
 }
--- a/tools/createVanillaPackageJson.js
+++ b/tools/createVanillaPackageJson.js
@@ -4,12 +4,20 @@ import path from 'path'
 import { readFileSync, writeFileSync } from 'fs'
 import { getAbsolutePath } from './getAbsolutePath.mjs'
 import { getFilesRecursively } from './getFilesRecursively.js'
+import { packageFolder, name, moduleFile as module, umdFile as main } from '../rollup.config.bundle.js'
 
-const vanillaPackageFolder = getAbsolutePath(import.meta.url, '..', 'package-vanilla')
+const vanillaPackageFolder = getAbsolutePath(import.meta.url, '..', packageFolder)
 
 // collect all file names and generate the exports map for package.json
 const exports = {
-  '.': './index.js'
+  '.': {
+    'browser': {
+      'import': './' + module,
+      'require': './' + main
+    },
+    'import': './' + module,
+    'require': './' + main
+  },
 }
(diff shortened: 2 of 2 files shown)


josdejong:
Thanks for your inputs Cloyd and thinking this through.

To be honest, I really would love the JavaScript world to move to ES modules and get rid of CommonJS asap. Adding UMD support feels like a step in the wrong direction, though it may (still) be necessary. Your support in `json-editor-vue` for for previous versions of Vue and Next is admirable on the one side. But on the other side: we as developers make our own lives much harder by keeping supporting outdated versions of everything for a long time. A bit of an incentive to keep your stuff upgraded may be good for all of us. I'm not a Vue expert, is the idea that the world moves to Vue 3, or do we get something similar to the Python 2/3 nightmare (I hope not)?

CommonJS or UMD support can still be necessary today, though I think we should not spend effort in getting the library to work on old browsers that do lack ESM support (they also lack other JS features). It feels like https://github.com/cloydlau/json-editor-vue/issues/5 can be solved without publishing an UMD bundle. If you want to use the ESM library as a "regular" script, you can do:

```html
<script type="module">
  import { JSONEditor } from 'vanilla-jsoneditor'
  window.JSONEditor = JSONEditor
</script>
```

So far, #165 is the only case of a user stuck with an "outdated" setup that cannot use the ESM library. If there are more compelling real-world cases I'm open to discuss adding UMD support but so far I prefer supporting ESM only (and having workarounds for _most_ cases of older setups).

cloydlau:
> we as developers make our own lives much harder by keeping supporting outdated versions of everything for a long time

I totally agree with this 👍, but the fact is that users are not always keeping up with the times and we are not supposed to disregardfully abandon them.

Vue 3.0 official has released **over 2 years** and still lag far behind 2.x on downloads:

image

It doesn't mean Vue 3 is not good, imaging if a website was built with Vue 2, it most likely won't upgrade to Vue 3 unless insurmountable problems arise.
For independent developers, there's concern about stability, prosperity of ecosystem and learning cost.
For enterprises, stability matters more in most cases.

Interesting thing is I'm writing Vue 2 in my company and Vue 3 in my own projects at the same time. It's actually not that bad thanks to the tremendous effort Vue team made on version 2.7, `@vue/composition-api`, `unplugin-vue2-script-setup`, `vite-plugin-vue2` & `@vitejs/plugin-vue2`, which narrowed the incompatibility between Vue 2 & Vue 3 to a very small degree. Vue 2 users benefit from those a lot, that's the time I realized backward compatibility is so valuable. If Vue team left Vue 2 users regrardless, we can say that's a Python 2/3 nightmare.
Similarly, `jsoneditor` is still under maintaining while `svelte-jsoneditor` is reaching its second birthday.

Conclusions:
1. **Time**. Backward compatibility is required, it's the time we should focus on. Imaging this:
The highly anticipated `:has()` pseudo-class grammar came out 2 months ago, we can't say let's use it and blame user on not catching up.
Native ESM is not supported by 5-years-ago Chrome, 4-years-ago Firefox & 3-years-ago NodeJS. Should 3 to 5 years be considered 'outdated'?
2. **Cost**. How much cost do we pay and is there sacrifice we need to make for it?
Seems no for this one.

> If there are more compelling real-world cases I'm open to discuss adding UMD support

No problem. Let's wait for more infomation.

cloydlau:
Actually there're two .cjs files in svelte-jsoneditor 😂:
- ./eslintrc.cjs
- ./commitlint.config.cjs

josdejong:
Yes agree, it is a balancing act between supporting users and having to maintain old stuff for a long period. Things should just work and keep working. I'm a big proponent of keeping things backward compatible and stable too, like you and I guess every developer. But that is not the same as keeping old,  _backward-incompatible_ versions alive for a very long period. Lately I'm becoming a bit more skeptical of that, since there is a big cost involved.

When will your company switch to Vue 3? I guess when there is some time left for a migration (which is never), or when old things _actually_ start to break, forcing an upgrade to Vue 3. When will things start to break? As soon as library developers stop supporting Vue 2. When will they do that? As soon as most users migrated to Vue 3. When are most users be migrated? -> go to the start of this paragraph. 

I'm afraid that in practice, companies only migrate after the EOF is reached and things start to break, no matter whether that is after a 2 or 5 year period. It is a matter of trading developer time of a company (postponing time needed to upgrade a project) against developer time of the frameworks/library developers. Now, for the Vue team it probably doesn't cost much effort to keep Vue 2 around. But there is most likely a huge amount of hidden effort required by all the library maintainers having to maintain two version of the library (for Vue 2 and Vue 3), and for all Vue developers having to live in a hybrid ecosystem.

Anyway, interesting discussion, but back to `svelte-jsoneditor`: an important distinction to me is whether there is a workaround to keep things working in an older setup or not. So far, the (most) issues I've seen getting `svelte-jsoneditor` to work in an older environment _do_ have a workaround. If there are enough people that can't use the library at all because there is no workaround, I'm ok with adding UMD support. Does that make sense?

And good point about the two `.cjs` files in the project, I'll change them into `.json`. Also, the `svelte-jsoneditor` project does have a couple of dependencies that are CommonJS. I would love to see those upgraded as well.

cloydlau:
Hahaha! The infinite loop theory is very interesting. I guess if all developers had same thoughts like you will make this industry a better place. The migration will happen anyway, a matter of time, smoothly or radically, with low or high cost. I plan to open a discussion to hear what people say. Feel free to close this, an instructive conversation is more valuable.

josdejong:
Ha ha. Well, I think all developers have good intentions and everyone wants to move forward, it's just that reality is so unruly and there's is no silver bullet in how to go about this, these are forces bigger than me and you😅. 

I just try to push back a bit against using UMD/CommonJS, but if we really need it we can add it (or just publish a separate npm package for it or so).

Ok will close this PR now, thanks 👍

Issue #204, opened 2022-12-09
Import failing in sveltekit


luke-hagar-sp:
https://github.com/josdejong/svelte-jsoneditor/blob/99df4b302d9d60479a2dcfffa669af9f9619e358/src/lib/components/modes/tablemode/ColumnHeader.svelte#L3

Hey! it looks like you are importing ts files by name, and in this example, it's causing failures on import after transpilation.

I'm not a typescript expert by any means, but I think in these examples you usually need to leave the .ts off prior to transpilation for TS files, and I think its resolves all that for you with the new .js name after transpilation.

I'm happy to submit a PR to fix this, just don't want to step on any toes

luke-hagar-sp:
https://github.com/josdejong/svelte-jsoneditor/blob/99df4b302d9d60479a2dcfffa669af9f9619e358/src/lib/components/modes/tablemode/ColumnHeader.svelte#L9

looks like the same issue here actually

josdejong:
Ah, you're right. Unfortunately the build tools do not complain about this 🤔 . Ideally, they should be automatically changed to `.js`, or at least we should get a compile or lint error or so.

A PR fixing the wrong imports would be very welcome, thanks for your offer.

luke-hagar-sp:
![image](https://user-images.githubusercontent.com/98849695/206720048-fae643a0-fd7f-4774-b33a-459c47b6dad5.png)
I get linting errors, do you know if there are more files like that?

josdejong:
Yes I get these linting errors too, using a different IDE though. In my case I get the suggestion to shorten the import by replacing it with `"$lib"`, but that is not a good idea, since that is the index file for the library, and results in circular dependencies. (I do not yet know if I can turn off that suggestion somehow).

luke-hagar-sp:
Alright, I made a pass and found a couple other small issues, I'm aiming to use this instead of monaco in a production application for making and receiving manual API calls, so I just want to help button everything up

luke-hagar-sp:
Please let me know if you have any questions or issues, I look forward to using this tool in my app :)

josdejong:
Fixed in `v0.11.2`

PR #205, opened 2022-12-09
Update ColumnHeader.svelte

luke-hagar-sp:
Removing specified ts extensions. Import paths should not end with .ts, as the paths are no longer valid after transpilation.

Should address #204, I am also searching for other instances

### DIFF
--- a/src/lib/components/controls/navigationBar/NavigationBar.svelte
+++ b/src/lib/components/controls/navigationBar/NavigationBar.svelte
@@ -29,7 +29,7 @@
   $: path = documentState.selection ? documentState.selection.focusPath : []
   $: hasNextItem = isObjectOrArray(getIn(json, path))
 
-  function scrollToLastItem() {
+  function scrollToLastItem(path) {
     setTimeout(() => {
       if (refNavigationBar && refNavigationBar.scrollTo) {
         const left = refNavigationBar.scrollWidth - refNavigationBar.clientWidth
@@ -42,7 +42,7 @@
   }
 
   // trigger scrollToLastItem when path changes
-  $: scrollToLastItem()
+  $: scrollToLastItem(path)
 
   function getItems(path: JSONPath): string[] {
     debug('get items for path', path)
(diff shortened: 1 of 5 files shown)


src/lib/components/controls/navigationBar/NavigationBar.svelte
```diff
@@ -42,7 +42,7 @@
   }
 
   // trigger scrollToLastItem when path changes
-  $: scrollToLastItem(path)
+  $: scrollToLastItem()
```
josdejong, line 45: The path actually needs to be there as argument. It is an ugly hack to "trigger scrollToLastItem when path changes".

I'm open to better solutions, but at least let's keep it as it is for the time being and address that in a separate PR.

josdejong (changes requested):
Looks good, thanks Luke! 

I made one inline comment, after that we can merge your PR.

src/lib/components/controls/navigationBar/NavigationBar.svelte
luke-hagar-sp: ahhhhhhhh, I understand, it's for the reactive trigger.

In this example, I think its better to have the unused param on the function.
I amended my PR

josdejong: 👍 Agree I'll adjust a couple of similar cases too, and add `// eslint-disable-next-line @typescript-eslint/no-unused-vars`  above it to make the linter happy.

luke-hagar-sp: Roger That

josdejong (approved):

josdejong:
The fix is published now in `v0.11.2`. It is literally Friday 17:00 at my place, always a good moment for a release 😂. 

Thanks again Luke.

luke-hagar-sp:
:D

Now that I am able to use this package in my project, I am BLOWN AWAY with how feature-rich it is. You have REALLY outdone yourself here.

josdejong:
Thanks, good to hear 😊 !

Issue #163, opened 2022-10-12
Feature-Request: onSelect


sruehl:
add a `onSelect(path: Path, value: any): void` similar to `onClassName(path: Path, value: any): string | undefined`.

Usecase: I have two editors open and when I select one element in the one editor I want to evaluate the path and value and use it to jump to another element in the second editor.

sruehl:
as a small addon to that: this feature is used already internally for the breadcrumb navigation (just not exposed).

josdejong:
Yes, good idea! Thanks for your suggestion Sebastian.

Two thoughts in this regard:

1. I think an `onSelect` callback will mostly be useful if there is a corresponding method and/or option to change the selection too.
2. We'll need to cater for two different types of selection: in `tree` mode and `text` mode, or come up with a unified way to describe selections in both modes.

sagIoTPower:
The pull request #229 **add updateSeletion to JSONEditor** addresses the feature to change the selection.
I would be happy if it could be merged into main.

Regards Christof

josdejong:
Thanks @sagIoTPower. I would like to first think through https://github.com/josdejong/svelte-jsoneditor/issues/163#issuecomment-1277397051 before merging your PR, we may want to go a different direction to cater for the three different modes. Do you have thoughts on a good approach to that?

sagIoTPower:
Before thinking about a concept for a harmonised approach, I have some observations/ questions:

1. What is idea behind a unified way to address selections in all modes? In other features the different modes diverge as well.  I see `scrollTo(path: JSONPath)` is implemented in `TreeMode` and `TableMode` but not in `TextMode`.
2. Although `scrollTo(path: JSONPath)` is implemented in `TableMode` but it is never called:
snippet from `src/lib/components/modes/JSONEditorRoot.svelte`

``` 
export function scrollTo(path: JSONPath): void {
    if (refTreeMode) {
      return refTreeMode.scrollTo(path)
    } else {
      // TODO: implement scrollTo for text mode

      throw new Error(`Method scrollTo is not available in mode "${mode}"`)
    }
}
```

3. If we want to harmonise the selection in all three modes, it would mean to harmonise:
`function setSelection(anchor, head)` -> TextMode
and
`function updateSelection(...)` -> TreeMode, TableMode.

Is my understanding correct?

josdejong:
Thanks for your input @sagIoTPower. 

[...]

PR #209, opened 2022-12-23
Add onselect

iliakap:


josdejong:
Thanks @iliakap for your PR, I'll look into it after the holidays

josdejong:
Thanks again for your PR @iliakap .

For reference: this PR will resolve #163.

Two questions:

1. Right now, the `onSelect` will work for `tree` and `table` mode, but not for `text` mode. I think we need to at least have a clear plan on how to implement support for selection for `text` mode: either a different method, or via an extra parameter or so. What are your thoughts on that?
2. Besides listening for changes in the selection, we'll also need a way to _change_ the selection, like I describe in https://github.com/josdejong/svelte-jsoneditor/issues/163#issuecomment-1277397051. What do you think?

josdejong:
Closing this PR now due to lack of response. Feel free to reopen if needed.

iliakap:
@josdejong, just saw ur comments, sorry for the late response :)
I wont reopen now since I'll only have time to look at it in a few weeks. 
with regards to your questions:
1. having hard time to remember, but I recall the text mode is just free text, I think it can be implemented using the [web selection api](https://developer.mozilla.org/en-US/docs/Web/API/Selection), but it felt out of scope.
2. I agree it's a nice addition, LMK if you wanna add this to this PR, not sure what the timeline will be though :)

josdejong:
Thanks for getting back @iliakap, no problem at all. Let's continue the discussion in #163

PR #240, opened 2023-03-20
ci: setup Release GitHub Action

cloydlau:
Hello,

How about setting up a Release GitHub Action?

Motivation:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Release Notification:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

- Wrapper author like me and whoever interested will immediately get notified when new versions released. I rely on this to ensure `json-editor-vue` catches up with the breaking changes of `svelte-jsoneditor` in time. It's very important to downstream ecology. (We can see 7 people are already watching)

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Without Release GitHub Action:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;With Release GitHub Action:

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;

- As we can see versions are displayed more clearly rather than hiding in tags or CHANGELOG.md.
- Don't worry about the release notes, [changelogithub](https://github.com/antfu/changelogithub) will do the job.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: latest
+
+      - run: npx changelogithub
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}


josdejong:
Just wondering about the PR itself: so it will not publish via Github, but only send out a message or so? I haven't seen this before.

cloydlau:
the PR itself: Yes it won't publish via GitHub, relax no refactoring will happen :), it won't impact the current release process, it's more like an add-on.
I personally write [scripts](https://github.com/cloydlau/scripts) for release process myself, so I can release my package by simply run `cl release`.
I'm not sure about dependabot. I think creating a PR is to heavy for that, my purpose is to get notified for releases of a specified package,  I think email is a good choice.

cloydlau:
Releases are based on [Git tags](https://git-scm.com/book/en/Git-Basics-Tagging), which mark a specific point in your repository's history:

image

josdejong:
👍 I'll merge your PR now, and somewhere this week I expect to publish a few small bug fixes so we can try it out.

Yeah for this project too there are also some scripts in place so you can do `npm run release` and that's it. I'm using [`standard-version`](https://www.npmjs.com/package/standard-version) there though to bump the version numbers, and this is now deprecated. Also, the scripts require to have the right stuff installed locally (node, git, npm), and you have to enter a code for two-factor authentication manually. So, an automated release in the cloud like with Github Actions would be beneficial instead, it would be more robust. But that's for a next step.

cloydlau:
Perfect. I‘ve received the v0.16.1 release notification email, and the release note works well too:

image

josdejong:
Yes indeed, https://github.com/josdejong/svelte-jsoneditor/releases/tag/v0.16.1

The release information includes two assets with zip files of the source code. I do not think that is useful (probably only confusing). Do you know whether it is possible to remove those?

josdejong:
Yes I see, I just think it's not very useful (the build output could be useful, but the source not really, is not usable as release). Anyway, not a big deal.

cloydlau:
I found Github official feedback for this: https://github.com/orgs/community/discussions/6003

josdejong:
Good find 👍

Issue #266, opened 2023-05-01
expand all fields in tree mode in ReactJS


nitzcard:
Hi, I try to expend all the fields in tree mode in react right from the start, 
I saw there is an expand function:
`refEditor.current?.expand((path) => true);`

I try to do it in the useEffect when initting the library but it doesn't work, 
do you have an idea how it can be achieved?

nitzcard:
https://github.com/josdejong/svelte-jsoneditor/pull/267

josdejong:
Thanks for your PR @nitzcard . I'm not entirely sure I understand the original issue: at the start, the document contents _is_ expanded by default. See for example this basic React example: https://codesandbox.io/s/svelte-jsoneditor-react-59wxz

nitzcard:
At my internal system it doesn't start expanded all the times,
I think it has to do with getDefaultExpand()

josdejong:
Can you share a minimal example demonstrating your issue?

The only case where I know it happens is if you initially load an empty document, and later _update_ it with actual contents. The editor leaves the expanded state as is when updating a document. When I load a new document, I normally create a new editor with clean state, so it expands the document contents like it should.

nitzcard:
Could be that's what happens at my form infra internally, when the data loads after. What do u suggest?

josdejong:
> If that is the case, how can I enforce expanded at the first time the actual data loads?

Either create a new editor with the document, or use `.set(...)` initially instead of `.update(...)`.

nitzcard:
I use `updateProps` as you wrote in your react demo to load the content, as I can't load it initially. 
To my understanding it doesn't creates a new editor so why it breaks the expanded state? 
This is a workaround, but it still not works 100% of the time: 
![image](https://user-images.githubusercontent.com/123297447/235626776-e3e881f8-f12b-4fda-b9f3-bb79ee705dcf.png)

can you help to think of a solution?

josdejong:
I think what should work is:

- do _not_ pass `content` via `refEditor.current.updateProps`, remove it from the props before passing
- when loading the document initially, use `refEditor.current.set(content)` (that will reset the state of the editor)
- when loading the document consecutively, use `refEditor.current.update(content)` (that will keep the state of the editor)

Alternatively, you can create the editor only after you have the document and can directly pass it during initialization of the editor.

nitzcard:
[...]

PR #267, opened 2023-05-01
added startExpanded prop

nitzcard:
Added the ability to expand fields right from the start in tree mode. 
I had a problem doing it in reactJS so I chose to contribute :)


josdejong:
Will close this PR now, see #266.

Issue #285, opened 2023-06-16
Add to CDN


KoalaBear84:
Hi Jos,

Thanks for your answer for pointing me to the successor. I would like it to be added to cdnjs.com, but I don't know what to put at the "Files to include", many times there is a "dist" folder with the files that are build, but this time I have no clue what files to include.

If you use this link, everything else will be filled in already.

https://github.com/cdnjs/packages/issues/new?assignees=&labels=%F0%9F%8F%B7+Library+Request&projects=&template=library-request.yml&title=Add%3A+svelte-jsoneditor&name=svelte-jsoneditor&license=https%3A%2F%2Fgithub.com%2Fjosdejong%2Fsvelte-jsoneditor%2Fblob%2Fmain%2FLICENSE.md&source=https%3A%2F%2Fwww.npmjs.com%2Fpackage%2Fsvelte-jsoneditor

Could you take a look? You can also create it yourself, otherwise I'll submit it after I know which files.

Thanks!

josdejong:
You can use for example jsdelivr:

https://www.jsdelivr.com/package/npm/svelte-jsoneditor
https://www.jsdelivr.com/package/npm/vanilla-jsoneditor

Is that sufficient?

KoalaBear84:
Maybe.. if I know how it can work on the regular web 😅

https://dev.to/iggredible/what-the-heck-are-cjs-amd-umd-and-esm-ikm

I think the umd thing should work on the regular web? But not sure if that is possible.

`Uncaught SyntaxError: Cannot use import statement outside a module (at index.min.js:7:1)`

Yes, I don't have any experience with all the module like techniques, only real plain JavaScript.

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>JSON Editor</title>
</head>
<body>
    <script src="https://cdn.jsdelivr.net/npm/svelte-jsoneditor@0.17.7/index.min.js"></script>
</body>
</html>
```

Also tried:
```html
<script type="module" src="https://cdn.jsdelivr.net/npm/svelte-jsoneditor@0.17.7/index.min.js"></script>
```

But got:
`Uncaught TypeError: Failed to resolve module specifier "immutable-json-patch". Relative references must start with either "/", "./", or "../".`

--

After that I tried the vanilla-jsoneditor, which I guess I should've tried the first time, but also have issues with that.

```html
[...]

Issue #265, opened 2023-04-29
Where is dist folder for javascript use


hoogw:
I want to use it in browser with plain javascript. 
But after I download .zip, can't find dist folder.

I don't know how to build to generate this file.

 /node_modules/svelte-jsoneditor/dist/jsoneditor.js 

Can you include dist folder in release .zip file or show us how to bundle it to generate a dist folder?

This is error when I try to build dist folder,

<img width="605" alt="Screenshot 2023-04-29 at 10 20 58 AM" src="https://user-images.githubusercontent.com/419582/235315631-0590760c-a0be-4589-815b-850bb6e570ad.png">

I also tried 
 ```
// svelte-jsoneditor (vanilla-jsoneditor for plain js)
   <script src='/public/utility/vanilla-jsoneditor/index.js'>
```
does not work either

I tried 53, but doesn't make it work for me
#https://github.com/josdejong/svelte-jsoneditor/issues/53

try #https://github.com/josdejong/svelte-jsoneditor/issues/165
still not work,

How to make it works?

josdejong:
I think you're looking for the vanilla bundle. This bundle is an ES module though, not UMD, see #165, #195.

Docs: https://github.com/josdejong/svelte-jsoneditor#standalone-bundle-use-in-react-vue-angular-plain-javascript-

hoogw:
I did a test page, but it does not work, I can't see it. However, it load ok, I don't see error in console. 

Here is code, 
<img width="1368" alt="Screenshot 2023-05-03 at 4 06 37 PM" src="https://user-images.githubusercontent.com/419582/236069651-5fd1dfcc-f7d5-466f-92a7-670e5f7f888e.png">


Here is f
<img width="718" alt="Screenshot 2023-05-03 at 4 08 20 PM" src="https://user-images.githubusercontent.com/419582/236070415-3de83665-98e6-491b-9017-d980d277e91b.png">
older

josdejong:
Do you get the source mapping in the following example? https://jsbin.com/yuwapaj/edit?html,output 

![afbeelding](https://user-images.githubusercontent.com/568626/236153145-2a16b84e-2d67-4206-8934-a291c0922b08.png)

hoogw:
Thank you,  [josdejong](https://github.com/josdejong) that solution is perfect !!
I closed this ticket. 


This is how I use new vanilla jsoneditor, click link

[...]

PR #286, opened 2023-06-16
Add CDN usage example

KoalaBear84:
If OK, please add/merge this. It will supply an example to the current standalone part to prevent 'issues' like #285 and #265

You can of course change whatever you like or make it more clear than my provided example.


josdejong:
Thanks, good idea!

I'll change the links to not include the version number (that will get outdated), and will include a warning better not to use a CDN in production in general.

Issue #318, opened 2023-10-05
Programmatic changes do not trigger 'onChange'


cloydlau:
Hello Jos,
The doc says 'The onChange callback which is invoked on every change of the contents, both changes made by a user and programmatic changes made via methods like .set(), .update(), or .patch()'.
But now it isn't, so is it intended or a bug?
Reproduction Link: https://codesandbox.io/s/svelte-jsoneditor-vue-forked-gp453r?file=/src/components/VueJSONEditor.vue
And another problem is that if I replace the `set` method in the Reproduction Link with `update`, the view won't update.

josdejong:
Thanks for reporting. The onChange in your example indeed doesn't work. Can you simplify the demo further?

I started from a bare bone JS demo testing `updateProps`, `update`, and `set`, and in that simple case `onChange` works fine with `updateProps`, `update`. It indeed looks like the method `set` isn't working at all. I'll look into that next week.

https://jsbin.com/jimehuh/edit?html,output

josdejong:
Can you strip the demo of the issue further so we can triangulate the issue?

Can you also update to use the latest version of the library? You're demo uses `v0.5.0`, the latest version is `v0.16.1`

josdejong:
The issue may be related to your example making a nested change in the existing `content` (mutating it). If you don't mutate but instead create and pass an updated object, the editor apparently doesn't always detect that the contents changed when you pass in the same object.

josdejong:
I've experimented with implementing support for mutating JSON objects. The editor is built around the assumption that the data is immutable. This has a number of big memory and performance advantages. Without immutability, the editor will have to make a full copy of the document with every change that is made, and it will have to re-render the full document as well. With immutable objects, it only has to render the changed parts and does not have to create deep copies.

What I can do is create an option `{ immmutable: boolean }`, and when `false`, let the methods `get()`, `set()`, `update()`, and callback `onChange` create a deep copy of the document. This will work fine with small documents. I think it would be best to set `immutable: false` by default (that is the safest approach). Does that make sense?

josdejong:
I've created a PR testing this out with a new option `{ immutable: boolean }` on the editor. The performance impact is really bad (deep copying the document on every change), and only needed when people do programmatic changes themselves outside of the editor. So I'm in doubt whether it is a good idea to set `{ immutable: false }` by default, giving everyone the slow-but-safe editor by default.

An other option would be to just document the current behavior and explain that you should not mutate your data.

[...]

PR #332, opened 2023-10-31
feat: create option `immutable`, implement support for mutable changes

josdejong:
See #318

TO DO:

- [x] Decide on whether to go for this option or not (think through an other API)?
- [x] Write docs
- [x] Fix broken unit test
- [x] Test extensively
- [x] Think through what other API we need to rework, for example support for mutating the selection object?
- [ ] Update all examples to configure `immutable={true}`
- [ ] Decide whether to go with this PR or not


josdejong:
We will not finish and merge this PR for the time being. The editor requires immutable changes, and this is now documented in the README, see 0545e647b9c0e68edecbaf7c8f65d51d64de99a4.

Issue #334, opened 2023-11-02
Running Jest doesn't find 'vanilla-jsoneditor' module


Rob-Negrete:
Hi @josdejong - 
I'm moving from the old jsoneditor to vanilla-jsoneditor.  It was pretty easy, and the component works on runtime as a charm. 

However, there's some weird issue when running test cases with jest.  

My implementation is basically the same as [JSONEditorReact.tsx](https://github.com/josdejong/svelte-jsoneditor/blob/main/README-VANILLA.md#typescript). Then, when I run tests, this error is thrown: 

```
 ● Test suite failed to run

    Cannot find module 'vanilla-jsoneditor' from 'JsonEditorReact.tsx'

    Require stack:
     JsonEditorReact.tsx

      at Resolver._throwModNotFoundError (../../../node_modules/jest-resolve/build/resolver.js:427:11)
```

I've installed `vanilla-jsoneditor": "^0.18.11`  and included as part of my dependencies. 

```
"dependencies": {
    "react": "^16.14.0",
     ....
    "vanilla-jsoneditor": "^0.18.11"
  },
```

Accordingly, to what I've read, this can be related to jest-resolver not being able to resolve the types / mapping.

Am I missing something?

Thanks in advance.

josdejong:
Thanks for reporting and working on a fix in #335. 

How can we reproduce this issue exactly? 

Do you have any pointers of other projects having similar issues with Jest? I would like to better understand what's going on.

josdejong:
> It is like if the module wasn’t being detected or something at least at testing time.

Indeed, and the question is why. How can I reproduce this issue?

Rob-Negrete:
I tried to reproduce it on a separated project (standalone without monorepo) but worked fine. So, I'll to set up a tiny monorepo project and try to reproduce the error on it. 
If I had success, I'll share it to you.

josdejong:
Thanks Rob!

Rob-Negrete:
I'd sent you an invitation to my repo which you can use it to reproduce the error.

josdejong:
Thanks for setting up the demo, I can reproduce the issue in your setup. It's quite a complex setup though 😅 .

I've done a bit of digging. The root cause seems to be that Jest does not have full ESM support (it is [still marked experimental](https://jestjs.io/docs/ecmascript-modules)).

[...]

PR #335, opened 2023-11-02
Fix exports for vanilla package

Rob-Negrete:
As described on https://github.com/josdejong/svelte-jsoneditor/issues/334

This attempts to make vanilla-jsoneditor module available, specially when working in monorepos.


josdejong:
Thanks for working on a fix for this issue Rob!

I'm in doubt whether this is a correct solution: the `package.json` file has configured `"type": "module"`, and the `index.js` file in the package is an ESM file. There is no a CommonJS file. It feels like a wrong solution to introduce an entry for `require` in the `package.json` pointing to an ESM file, a `require` entry should point to a CommonJS file (but there is none, so that doesn't make sense) 🤔.

Rob-Negrete:
Hi @josdejong,

On the contrary, thanks for making some time to review this. 

Yeh, I agree there's should be a better solution than this. I tried several alternatives and this one was the first that worked to me.

Rob-Negrete:
I've updated the PR. I've compared against @tinihttp/app's package.json lib. Which is also a pure ESM and it works fine, so I gave it a try and use the same `exports` declaration. 

Not sure if this makes sense or not, but hopefully it does.

josdejong:
Thanks for paving the path and debugging this issue Rob! I've worked out the solution that you've found in https://github.com/josdejong/svelte-jsoneditor/commit/3058c66c6536eaf2902f799a2e59054dbb7d4df2, which is now published in `v0.19.0` Closing this PR now.

PR #341, opened 2023-11-21
Update ValidationErrorsOverview.svelte

hybridwebdev:
Error message output component lacks much needed classes.
Added additional classes for improved CSS targeting.


josdejong:
Thanks @hybridwebdev , this is a neat improvement!

Can your run `npm run format` to fix the linting issues? (should use double quotes in HTML)

hybridwebdev:
Added commit to use double quoted class names

josdejong:
👌

PR #343, opened 2023-11-23
Adding onHover event

TimHeckel:
This is a simple update to add an `onHover` custom event - I'd like to be able to optionally provide tooltips or something else when users hover over a node.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules
 /package-vanilla
 /.svelte-kit
 coverage
+.nvmrc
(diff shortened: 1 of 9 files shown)


josdejong:
Thanks for your PR Tim!

I think there is a bigger challenge here then just introducing an "onHover" (mousedown). In the same way, people may need a trigger on mouseup, mousemove, mouseover mouseout, etc, The reason that you can't add a listener on the root DIV element is the `event.stopPropagation()` that is used in some of the event listeners, to prevent the event from triggering recursively. Maybe we can solve that in a different way: somehow add a flag to the event or store the event internally, so that the nested JSONNode component knows the event is already handled by a child component. What do you think?

josdejong:
Ping @TimHeckel any thoughts?

josdejong:
I'll close this PR now due to a lack of response. Let's first discuss what the most suitable solution for this issue will be.

PR #353, opened 2023-12-06
Feat/external dependencies

josdejong:
This will change the `vanilla-jsoneditor` npm package such that it does not embed dependencies but instead import them. So, if you use these dependencies (like `Ajv`, `lodash-es`, `jsonrepair`, etc) too in your project, they can be shared instead of being bundled twice.


josdejong:
published now in `v0.20.0`

PR #354, opened 2023-12-11
Documentation link fix.

osverdlo:
Fix: The link to the full list with all variables was broken


josdejong:
Thanks Oleg!

josdejong:
O wow, after merging I see we also need to rename the extension from `.css` to `.scss` 😂 , I will do that right now.

Issue #393, opened 2024-01-22
Transform modal: Return javascript query from modal (and populate modal with existing query)


conorw:
This control is exactly what I was looking for... with one exception:

Is it possible to expose the JavaScript query that is generated in the transform modal?
Likewise, is it possible to populate the transform modal with an existing query?

conorw:
apologies @josdejong , I should have explained better:
Ideally, I want to be able to save the javascript transformation code that is built in the transformation modal dialog.
Hopefully the following quick and dirty PR should explain better https://github.com/josdejong/svelte-jsoneditor/pull/394

josdejong:
Ah, yes, so you want to be able to programmatically initialize the query when opening the transform modal, and, I guess, get a callback with the applied query when the user clicks "Transform" in the modal. I like to open up the API for this!

We have to think through what API we want exactly. Some first thoughts:

- I think we'll have to implement an API for both the Transform modal and the Sort modal to be feature complete
- Besides the query field, the Transform modal also has a wizard. I think we should be able to set all off the state, not just `query`.
- We could introduce properties like `query` and `onQuery`. Or maybe we should introduce hooks like `onTransformOpen` and `onTransformClose`, allowing to get and set the query and all other state.

PR #394, opened 2024-01-22
feat: return transform query

conorw:
expose the query property and return when transform has been amended


josdejong:
Closing this PR, discussion is still open in #393 (this PR served as an illustration, see https://github.com/josdejong/svelte-jsoneditor/issues/393#issuecomment-1904841142)

PR #396, opened 2024-01-23
feat: force context menu

1pone:
With the release of `onRenderContextMenu`, users have the opportunity to define a variety of menus based on their needs, which may be more than just editing-related, and which are currently not presented when in `readOnly` mode.

### DIFF
--- a/src/lib/components/JSONEditor.svelte
+++ b/src/lib/components/JSONEditor.svelte
@@ -75,6 +75,7 @@
   export let askToFormat = true
   export let escapeControlCharacters = false
   export let escapeUnicodeCharacters = false
+  export let forceContextMenu = false
   export let flattenColumns = true
   export let parser: JSONParser = JSON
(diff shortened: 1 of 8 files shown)


josdejong:
Thanks, good idea. Right now the context menu is simply not rendered when read only.

I was thinking. When a user sees this new property `forceContextMenu`, it may not be clear what it exactly is and that it is related to a combination of `onRenderContextMenu` and `readOnly`. Maybe we can instead extend `onRenderContextMenu`: always render it (also when readOnly), unless the function returns `false` for example. We can pass `readOnly` along with the `context` object (2nd argument). The default implementation of `onRenderContextMenu` can return `false` when `readOnly` is `true` to maintain the current behavior. What do you think?

1pone:
You make a good point, the new property `forceContextMenu` introduced will make the logic more difficult to understand. Maybe I'll try the solution you're talking about.

Issue #375, opened 2023-12-21
How to define RenderValueComponent while using vanilla-jsoneditor


1pone:
I'm using vanilla-jsoneditor in React, and i need use custom RenderValueComponent  for `onRenderValue.component` to render value as password, like the ReadonlyPassword example shown in  the project. But how can i define the RenderValueComponent without svelte?

1pone:
Or can I submit a PR to add `ReadonlyPassword` export?

josdejong:
I haven't tried out if/how you can use `onRenderValue` from an other framework like React, we have to do some experimentation there.

I like the idea of exporting `ReadonlyPassword`: it can be the first a collection of extra value render components. I think though that we should either rename it to something like `HiddenValue`, or improve on it: it was just a minimal example, and when you double click it to edit, you see all of the password in plain sight, that should remain hidden whilst typing.

AlexRMU:
I think you can do it like [this](https://svelte-5-preview.vercel.app/#H4sIAAAAAAAACn1TwY6cMAz9FSvaA6hopr0yw0j9hR56aaoKgplmFxyamO2uEP9eJWGYoNGWC-C8Zz8_O7PodI9OlD9mQfWAohRfx1EUgt9H_-NesWcUhXBmsspHJJ-dsnrkiyQAAD2MxjLMYOgbUov2e91PCAt01gwgxeGoqcW3w7OT4hQpPTIoMxFDBU-Oa8bsc36SFE-7iRRrQ6BJWRyQOMthjmf-icxPFXw5pUFypsdDb66ZFBtTiiLi8xW73Kp4DaM1o_MaWrT6FdtsjuACateWIMWfvygFLHmqu4EKshyqC8zLGn_CrkPF2RpOVDVZqLFVv30kPeqH9sgxKIs1I1R7V_fZ7mhDv9Tvmq6eEJlZa9TkHTg0pn3fMZqQdSXcbTkft7FKOjcTsyEwpHqtXqo5tpZMZLmEE1fCHDxbzsfIufyHrQPRvxO0KMRgWt1pbEXJdsKl2HYxrM59G59duombiVs3qz83O_dLsaH8Uszr8MNEwlJs2VYDubZX5A9yRUxMFIH32VrkydLO4lAA38JFSVQ_TvYj5QnyQX1SM8ryBR9N_bn8AwuB7aXsAwAA) (after https://github.com/josdejong/svelte-jsoneditor/issues/373).
[...]

PR #397, opened 2024-01-23
feat: hidden value plugins

1pone:
Export `HiddenValue` value plugins, to fix the inability to customize value components for version vanilla

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -3,6 +3,7 @@ import BooleanToggle from './plugins/value/components/BooleanToggle.svelte'
 import ColorPicker from './plugins/value/components/ColorPicker.svelte'
 import EditableValue from './plugins/value/components/EditableValue.svelte'
 import EnumValue from './plugins/value/components/EnumValue.svelte'
+import HiddenValue from './plugins/value/components/HiddenValue.svelte'
 import ReadonlyValue from './plugins/value/components/ReadonlyValue.svelte'
 import TimestampTag from './plugins/value/components/TimestampTag.svelte'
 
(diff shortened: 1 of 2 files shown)


src/lib/plugins/value/components/HiddenValue.svelte
```diff
@@ -0,0 +1,45 @@
...
+      // open in edit mode
+      onSelect(createValueSelection(path, true))
+    }
+  }
+</script>
+
+<div
+  role="button"
+  tabindex="-1"
+  class="jse-value jse-readonly-password"
```
josdejong, line 29: Can you rename the class to `jse-hidden-value`?

src/lib/index.ts
```diff
@@ -13,7 +14,15 @@ export * from './types.js'
 // value plugins
 export { renderValue } from './plugins/value/renderValue.js'
 export { renderJSONSchemaEnum } from './plugins/value/renderJSONSchemaEnum.js'
-export { BooleanToggle, ColorPicker, EditableValue, EnumValue, ReadonlyValue, TimestampTag }
+export {
```
josdejong, line 17: Can you also mention this new component in the README?

https://github.com/josdejong/svelte-jsoneditor/blob/main/README.md?plain=1#L286

josdejong (commented):
Thanks Jesse! 

I made two small inline comments, can you have a look at those?

josdejong:
I had a look into the long term solution that is discussed in #375 to make it possible to create components in vanilla JS, and it turns out using Svelte actions works like a charm, see #398.

If we have that solution working, this `HiddenValue` component is probably not needed: you can implement it yourself then (and refine the experience to make it a true password component). What do you think?

1pone:
Wonderful! I'll try to implement it by myself in vanilla JS as soon as it's officially released. At that point, I'll close this pr if it works.

josdejong:
I've finished and merged #398 (not yet published)

PR #399, opened 2024-01-30
Feat: onRenderContextMenu return false

1pone:
Always hide the context menu when `onRenderContextMenu` return `false`

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,10 +333,11 @@ const editor = new JSONEditor({
       }
       ```
 
-  - `onRenderContextMenu(items: ContextMenuItem[], context: { mode: 'tree' | 'text' | 'table', modal: boolean, selection: JSONEditorSelection | null }) : ContextMenuItem[] | undefined`.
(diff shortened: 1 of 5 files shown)


1pone:
https://github.com/josdejong/svelte-jsoneditor/pull/396

josdejong:
Thanks! I think the approach to extend `onRenderContextMenu` to return `false` is the right approach. However I think implementing it neatly requires a bit more refactoring internally. The `onRenderContextMenu` callback should be invoked only once, and we should not introduce a "test call" with empty items to see if it does return false (that can lead to unexpected behavior.

Right now, `onRenderContextMenu` is called inside `TreeContextMenu` and `TableContextMenu`. I think this call should be pulled out of those components, it should be moved to the two `openContextMenu()` implementations: invoke `onRenderContextMenu`, if it returns false, cancel opening the modal, and otherwise, pass the items to the modal. Does that make sense?

1pone:
Thanks for the suggestion, I've tried to rewrite this part of the code in https://github.com/josdejong/svelte-jsoneditor/pull/411, can you help me see if there's any problem?

josdejong:
Published now in `v0.23.0`

PR #411, opened 2024-03-04
feat: hide context menu when onRenderContextMenu return false

1pone:
This is an optimized version of pr https://github.com/josdejong/svelte-jsoneditor/pull/399


josdejong:
Thanks! I think this is exactly the logic we need.

I have a few feedbacks regarding the code:

1.   I think it would be nice to keep the logic to create the menu items in a separate file, since the files `TreeMode.svelte` and `TableMode.svelte` are already very large. We can create a function like `const items = createTreeContextMenuItems({ json, documentState, parser, ..., onCut, onCopy, onPaste, ... })` and put that in a separate file. Also, the components `TreeContextMenu` and `TableContextMenu` are now about empty and don't add anything anymore, I think we can remove them. So basically: we can replace the component `TreeContextMenu.svelte` with a function in a file `createTreeContextMenuItems.ts`, mostly containing the same logic as the component had (and do the same with table mode). What do you think?
2.  Minor thingy: can you replace the inline return like `if (items === false) return` to have braces, like:

    ```js
    if (items === false) {
      return
    }
    ```
    (a personal preference)

1pone:
Thanks, I've made the changes you suggested.

src/lib/components/modes/tablemode/TableMode.svelte
```diff
@@ -921,11 +923,10 @@
     offsetLeft,
     showTip
   }: AbsolutePopupOptions) {
-    const props = {
+    let defaultItems: ContextMenuItem[] = createTableContextMenuItems({
```
josdejong, line 926: Can you define `defaultItems` as `const`?

src/lib/components/modes/treemode/TreeMode.svelte
```diff
@@ -1020,7 +1022,7 @@
     handleInsert(type)
   }
 
-  function handleConvert(type: 'value' | 'object' | 'array') {
+  function handleConvert(type: InsertType) {
```
josdejong, line 1025: Thanks for extracting this in a type. Can you create a `ConvertType` here instead of using `InsertType`? The function doesn't accept `'structure'`, so it slightly differs.

src/lib/components/modes/JSONEditorRoot.svelte
```diff
@@ -122,7 +122,10 @@
   $: handleRenderContextMenu = (items: ContextMenuItem[]) => {
     const itemsOriginal = cloneDeep(items) // the user may change items in the callback
 
-    return onRenderContextMenu(items, { mode, modal: insideModal, selection }) || itemsOriginal
+    return (
+      onRenderContextMenu(items, { mode, modal: insideModal, selection }) ??
+      (readOnly ? false : itemsOriginal)
```
josdejong, line 127: 👌 this keeps the logic backward compatible

josdejong (changes requested):
Niice job!! This looks really good, and keeps the behavior backward compatible.

I made two small inline comments, can you have a look at those?

1pone:
Resolved

josdejong (approved):

josdejong:
Thanks!

I was thinking: this still has to be published as a breaking changes: when you had an `onRenderContextMenu` before and was returning something, you now have to handle the case when the editor is `readOnly`.

1pone:
That's right, just like the adjustments made in the [development page](https://github.com/josdejong/svelte-jsoneditor/pull/411/files/f04418a14d88cfbbb767a586463313f47b56cad0#diff-347d85ce8eb5e9f882f205e5b4a9b7d0085a14a74ff9abdabd8128e58b6da6e4L383-R383), this should really be known by those who are or have used `onRenderContextMenu`.

josdejong:
I had one more thought: now that the context menu can be visible when the editor is readOnly, all the buttons should reckon with it and be disabled when not applicable in read-only mode. I've worked that out in c66ee09. I also came a cross a few minor issues related to the context menu, fixing that too.

Also, I added a property `readOnly` to the context of the `onRenderMenu` and `onRenderContextMenu` callbacks, see 4df55481a17c22211538df1cd289faec3af25177.

Issue #150, opened 2022-09-29
custom validator severity


Reload0213:
I'd like to ask you one more question. The [example](https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/examples/custom_validation/%2Bpage.svelte) you 
example provided in the last question made it possible to use a custom validator.  

However No matter what value is delivered as a key value of severity to the validation of the currently customized guidance message, it is output like the yellow "warning" format. As before, I want to pass values ​​like "info", "warning", "error" etc to isolate fatal errors and make them noticeably reddish.

I wonder if I should deliver a value other than "info", "warning", and "error" to the severity key of the customized inspector or not.

josdejong:
That is a good suggestion indeed, it would be nice to be able to specify the severity coming out of the validator.

Right now, the data models already have a `severity` field defined, but the UI doesn't utilize this information:

https://github.com/josdejong/svelte-jsoneditor/blob/f70a602d24eff3a57312857a61626ca53a1b2456/src/lib/types.ts#L154-L164


So we should reflect the severity in the rendered validation error table and the inline warning icons (both in `text` and `tree` mode).

Help implementing this would be welcome!

Reload0213:
If the ui reflects this point, it will be of great help in using the inspection.

Thank you for listening to me.

josdejong:
Moving this idea to the [Discussions > Ideas](https://github.com/josdejong/svelte-jsoneditor/discussions/categories/ideas) section since we are not actively working on it. Help implementing this would be welcome.

Issue #148, opened 2022-09-26
How can I change validation message?


Reload0213:
hi, Love your work!

Can I change validation message?

Even though strict mode is turned off, errorMessage and errorMessages of the ajv schema structure are not applied

please help me Thanks!

josdejong:
Thanks. Can you share a minimal code example of what you're trying to do?

Reload0213:
I want to customize the validation message as desired, or if not, apply the locale where I live.

So, knowing that validation is handled through ajv, I tried the following with reference to it.

`const SvelteJSONEditor = ({ props, refEditor }) => {
    const theme = useTheme();

    const refContainer = useRef(null);

    // validation(예시) - 필요한 알림 값 정의 후 적용
    const schema = {
        title: "Employee",
        description: "Object containing employee details",
        type: "object",
        properties: {
            firstName: {
                title: "First Name",
                description: "The given name.",
                examples: ["John"],
                type: "string",
            },
            lastName: {
                title: "Last Name",
                description: "The family name.",
                examples: ["Smith"],
                type: "string", // type 컬럼 타입 설정
                errorMessage: {
                    type: ":lastName:이름을 입력해 주세요",
                },
            },
            gender: {
                title: "Gender",
                enum: ["male", "female"], // enum 컬럼 값 제한
            },
            availableToHire: {
                type: "boolean",
                default: false,
            },
            age: {
                description: "Age in years",
                type: "integer",
                minimum: 0,
                examples: [28, 32],
            },
            job: {
                $ref: "job",
            },
        },
        required: ["firstName", "lastName"], // required 필수 컬럼 설정
        errorMessages: {
            // Change from errorMessage to errorMessages
            type: "should be an object",
            required: {
                firstName: "testesttest1111",
            },
        },
    };

    const schemaDefinitions = {
        job: {
            title: "Job description",
            type: "object",
            required: ["address"],
            properties: {
                company: {
                    type: "string",
                    examples: ["ACME", "Dexter Industries"],
                },
                role: {
                    description: "Job title.",
                    type: "string",
                    examples: ["Human Resources Coordinator", "Software Developer"],
                    default: "Software Developer",
                },
                address: {
                    type: "string",
                },
                salary: {
                    type: "number",
                    minimum: 120,
                    examples: [100, 110, 120],
                },
            },
        },
    };

    const ajvOptions = { allErrors: true, $data: true };

    useEffect(() => {
        // create editor
        try {
[...]

PR #416, opened 2024-03-14
Feat: Pass validation severity into UI

anlinguist:
This PR addresses the following discussion and issues:
- https://github.com/josdejong/svelte-jsoneditor/discussions/369
- https://github.com/josdejong/svelte-jsoneditor/issues/148
- https://github.com/josdejong/svelte-jsoneditor/issues/150

Allows the validation error severity to be `warning` (yellow), `error` (red), or `info`(blue).

If the validation error table is collapsed, it uses the first validation error to determine the color to show.
If the validation error table is opened, each validation error will be the assigned color from the validator.
Also updated the hover color for each validation error.


josdejong:
Thanks Andrew! I'll review your PR next week.

src/lib/components/controls/ValidationErrorsOverview.svelte
```diff
@@ -36,7 +36,7 @@
         <tbody>
           {#each limit(validationErrors, MAX_VALIDATION_ERRORS) as validationError, index}
             <tr
-              class="jse-validation-error"
+            class="jse-validation-error {validationError.severity}"
```
josdejong, line 39: I think it would be good to have a `jse-` prefix to prevent conflicts with classes from other frameworks like Bootstrap. (and similar to how you solved this in `ValidationErrorIcon.svelte`. 

However it makes sense to have a base class like `jse-validation-error`, and then three subclasses for the different severity, like `jse-validation-severity-error`, `jse-validation-severity-warning`, `jse-validation-severity-info`. These subclasses only need to alter the color. And then for consistency we can use the same class names in both the overview and the error icon. What do you think?

src/lib/components/modes/textmode/TextMode.svelte
```diff
@@ -643,7 +643,7 @@
       from,
       to,
       message,
-      severity: ValidationSeverity.warning,
+      severity: validationError.severity,
```
josdejong, line 646: You can use destructuring here to write this more compact, like `path` and `message` that come from `validationError` too:

```js
use destructure `const { path, message, severity } = validationError`
```

src/lib/components/controls/ValidationErrorsOverview.svelte
```diff
@@ -77,7 +77,7 @@
         </tbody>
       </table>
     {:else}
-      <table class="jse-validation-errors-overview-collapsed">
+      <table class="jse-validation-errors-overview-collapsed {validationErrors[0].severity}">
```
josdejong, line 80: Thinking aloud here: it would be nice to take the most severe severity here instead of the first. But this may be overengineering. What do you think?

src/lib/components/controls/ValidationErrorsOverview.scss
```diff
@@ -56,3 +54,18 @@
     }
   }
 }
+
+.error {
```
josdejong, line 58: Can you update the example [`src/routes/examples/custom_validation`](https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/examples/custom_validation/+page.svelte) to demonstrate the tree different severity levels?

src/lib/components/controls/ValidationErrorsOverview.scss
```diff
@@ -16,7 +14,7 @@
       cursor: pointer;
 
       &:hover {
-        background-color: rgba(255, 255, 255, 0.1);
+        filter: brightness(110%);
```
josdejong, line 17: This works perfect for the `error` and `info` classes, but in case of `warning` the color difference is very large. Is it possible to have a bit more subtle color change in that case?

josdejong: I also noticed that the highlighting doesn't work when the overview is collapsed, can you have a look at that?

josdejong (changes requested):
Thanks a lot for working out this solution Andrew, this looks good. I made a couple of inline comments. Can you have a look at those?

src/lib/components/controls/ValidationErrorsOverview.scss
anlinguist: I can set the brightness to 105% on `jse-validation-warning`. Also, migration to use `jse-validation-` prefixes has fixed the issue where hover brightness wasn't applied to the collapsed error.

src/lib/components/controls/ValidationErrorsOverview.svelte
anlinguist: makes sense. pushing a change soon that implements this

src/lib/components/controls/ValidationErrorsOverview.svelte
anlinguist: that's a good idea. also, instead of applying the class change to the table, I've changed it to apply to the row, and I'm taking advantage of the new class with `jse-validation-` prefix.

anlinguist: This also required a helper function `getValidationClass()`

josdejong: Nice 😎

src/lib/components/controls/ValidationErrorsOverview.scss
josdejong: Thanks, the 105% looks nice!

josdejong (approved):

josdejong:
Thanks for the updates, this looks really good, both the code and the UI!

josdejong:
I see there is a linting issue, I'll fix that right after merging.

josdejong:
After merging I thought it would be neat to rename the helper function `getValidationClass` to `getMaxSeverity`, I've changed that in 2c88d93648b25b193f9052ae365d9cd7152c0fa6.

josdejong:
Published now in `v0.23.1`, thanks again!

PR #426, opened 2024-04-24
fix: #424 resetting properties to undefined does not restore it's default value

josdejong:
See #424 

Todo:

1. Decide on whether this is the best approach
2. Think through whether we want to remove support for `null` in favor of `undefined` for the properties `selection`, `validator`, and `onSelect`


josdejong:
Closing this PR, better to look into a different direction for a solution. See https://github.com/josdejong/svelte-jsoneditor/pull/424#issuecomment-2116974777

Issue #454, opened 2024-06-24
Allow custom editable value renderers


VincentK-ZA:
Looking at the [custom value renderer](https://github.com/josdejong/svelte-jsoneditor/blob/main/examples/browser/custom_value_renderer.html) examples, none of them really take over the the editing rendering.

I've been working on getting this work, but from I could see, I had to fork the project in order to get it working.

Here's an example of a custom editable div in our project, using codemirror:
![custom-editable-div](https://github.com/josdejong/svelte-jsoneditor/assets/8651835/a16edb13-ecab-4bd9-a723-da7152bb3ff3)

Please see [this PR](https://github.com/stubber/stubber-svelte-jsoneditor/pull/2) (internal to the forked repo for now) for what I believe would need to be changed.  
This includes an EditableCodeMirror component for demo purposes on the examples page. 
Our actual implementation relies on `getContext` to get actual completion/suggestion data, since I didn't want to pass any more props through EditableValue.

Let me know what you think about this proposal, and if you'd be open to merging a change like this into the project.

There are some tricky things regarding focus/mousedown events that I have not managed to solve as of yet.  
For example, I couldn't get the suggestions panel to stay open if the user tries to drag the scrollbar to see more suggestions.

josdejong:
Thanks for your inputs Vincent.

Can you explain what you think is missing in API in order to (easily) create an editable value component yourself?

josdejong:
I've added an example showing how to create custom editable input:

- https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/examples/custom_value_renderer2/%2Bpage.svelte
- https://github.com/josdejong/svelte-jsoneditor/blob/main/src/routes/components/EditableValueInput.svelte

VincentK-ZA:
Hi Jos

Thanks for your reply and for the examples.

In our case, we are trying to make a custom editable input that functions and looks almost identical to the default editable div, just with autocompletions added.

So there are some internal/not exported utility functions that would be nice to reuse as is.
Namely: `getFocusPath, getValueClass, isObjectOrArray, keyComboFromEvent` we are currently relying on, and they are not currently exported it seems.

[...]

PR #461, opened 2024-07-10
feat: export more utility functions

josdejong:
See #454

### DIFF
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -74,6 +74,7 @@ export { onEscape } from './actions/onEscape.js'
 // type checking
 export {
   valueType,
+  stringConvert,
   isObject,
   isObjectOrArray,
   isBoolean,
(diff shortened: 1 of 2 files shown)


src/lib/index.ts
```diff
@@ -62,5 +71,16 @@ export { parseJSONPath, stringifyJSONPath } from './utils/pathUtils.js'
...
 
+// type checking
+export {
+  valueType,
+  isObject,
+  isObjectOrArray,
+  isBoolean,
+  isTimestamp,
+  isColor,
+  isUrl
```
VincentK-ZA, line 82: Can we also export [stringConvert](https://github.com/josdejong/svelte-jsoneditor/blob/75fd80cb8a7eea92fd0027766f92650eb4a0831f/src/lib/utils/typeUtils.ts#L163)

PR #490, opened 2024-10-15
Upgrade to Svelte 5

josdejong:
Issues:

- [x] [IMPEDIMENT] Incompatible dependencies: `svelte-awesome` and `rollup-plugin-svelte`. Solution: the dependencies must update their peer requirements to include Svelte 5. See https://github.com/RobBrazier/svelte-awesome/issues/1186#issuecomment-2420805965.
- [x] [IMPEDIMENT] Circular dependencies reported by Rollup, when running `npm run build`. Not a show stopper. Is a known issue, see https://github.com/sveltejs/svelte/issues/10140. Since this is not critical but just annoying, will mark this as solved for now.
- [x] New compiler warnings: "Invalid tr inside table". Solved in the `main` branch
- [x] New compiler warnings: "Self-closing HTML tags for non-void elements are ambiguous". Solved in the `main` branch
- [x] Compiler warnings "Unused CSS selector". A few valid cases are resolved in the `main` branch. The main cause is the recursive `JSONNode` component. It looks like in Svelte 5, the styles are not recursively applied to child instances of `JSONNode`. Solved by defining the recursive styling needed for nested child components using `:global`. It would be good to look into an alternative solution that does not require using `:global`.
- [x] Compiler warnings "Properties of objects and arrays are not reactive unless in runes mode". They mostly fire wrongly on TypeScript enum objects. Solved via a workaround by adding a temporary `warningFilter` in the svelte config. I think this is a bug in the svelte 5 compiler. UPDATE: this workaround is not suitable: when used as a svelte library, this requires the consumer to configure these warningFilters too, but that is not workable.
- [x] Some unit tests are broken. Fixed by mounting via `mount`, adding some config in `vitest.config.js`, and adapting the expected output and snapshots.
- [x] Scrolling by pressing the middle mouse button throws an error “Uncaught TypeError: handler is undefined”. Solved in the `main` branch by removing `on:focus={undefined}` and `on:blur={undefined}`, and instead ignore the svelte warning `a11y-mouse-events-have-key-events`. 
- [x] Styling of the vanilla output is broken. Solved by configuring `postcss` in the Rollup config.
- [x] Update `peerDependencies` to support svelte 5 and its release candidates 
- [x] Double clicking a value in tree or table mode doesn’t open edit mode for the value. Double clicking a key works fine. Looks like changing the selection in the mousedown handler in `JSONNode` breaks handling of the double click event.
- [x] Method `updateProps` of the vanilla library not working: `this.$set(props)` is removed in Svelte 5. 
- [x] Method `destroy` of the vanilla library doesn't work: `this.$destroy()` is deprecated in Svelte 5.
- [x] Update all dependencies to the latest version, change sass version again to use `^` and resolve all deprecation warnings.
- [x] The class `JsonEditor` is no longer exported by Svelte 5. Find an alternative solution

TODO:

- [x] Fix all open issues
- [x] Do a through functional test round. 
- [x] Test performance. It looks like the editor is about twice as slow on Firefox, and four times as fast on Chrome 🤔 
- [x] Test using the library in a Svelte 4, Svelte 5, vanilla, and React (TS) project
- [x] Test whether it is possible to keep the library compatible with both Svelte 4 and 5
- [x] Await the release of Svelte 5 and support for Svelte 5 by the dependencies
- [ ] Merge and publish


josdejong:
Ok all is ready 🎉 . 

Everything still works in Svelte 4 (and 3), which is amazing, and there should be no breaking changes at all. I'll still publish this as a major breaking change since there may be subtle edge cases (just like it this PR itself involved fixing some subtle breaking changes).

PR #522, opened 2025-02-17
feat: add createAjvValidatorAsync

iStefo:
This PR adds a new function, `createAjvValidatorAsync` to the Ajv plugin.

It behaves like `createAjvValidator`, but uses `ajv.compileAsync` on the schema which allows for resolution of remote schemas through the `loadSchema(uri)` hook of Ajv's `options`.

Due to the async nature of `compileAsync`, the return format changes, which would be a breaking change. So I opted for a second factory function in order to not break existing consumers.

The PR also adds `errorSeverity` as an option.

As you can see, I have refactored the code a bit s.t. as little code as possible is duplicated. It had a bit of an unfortunate impact on the testing code, where I also tried to not duplicate all the existing tests just be cause there now is a second factory function.

Thanks for your work & please let me know if you want anything changed :)

### DIFF
--- a/src/lib/plugins/validator/createAjvValidator.test.ts
+++ b/src/lib/plugins/validator/createAjvValidator.test.ts
@@ -1,8 +1,8 @@
 import { test, describe } from 'vitest'
 import assert from 'assert'
 import Ajv from 'ajv'
-import { createAjvValidator } from './createAjvValidator.js'
-import type { JSONSchemaDefinitions } from '$lib/types'
+import { createAjvValidator, createAjvValidatorAsync } from './createAjvValidator.js'
+import { ValidationSeverity, type JSONSchemaDefinitions } from '$lib/types'
 
 const schema = {
   title: 'Employee',
@@ -83,9 +83,20 @@ const invalidJson = {
   }
 }
 
-describe('createAjvValidator', () => {
-  test('should create a validate function', () => {
-    const validate = createAjvValidator({ schema, schemaDefinitions })
+describe.each([
+  { name: 'createAjvValidator', create: createAjvValidator },
+  { name: 'createAjvValidatorAsync', create: createAjvValidatorAsync }
+])('$name', ({ create }) => {
+  const options = {
+    schema,
+    schemaDefinitions,
+    ajvOptions: {
+      loadSchema: () => Promise.resolve({})
+    }
+  }
+
+  test('should create a validate function', async () => {
+    const validate = await create(options)
 
     assert.deepStrictEqual(validate(invalidJson), [
       {
@@ -99,12 +110,12 @@ describe('createAjvValidator', () => {
     ])
   })
 
-  test('should pass additional Ajv options', () => {
-    const validate = createAjvValidator({
-      schema,
-      schemaDefinitions,
+  test('should pass additional Ajv options', async () => {
+    const validate = await create({
+      ...options,
       ajvOptions: {
-        allErrors: false
+        allErrors: false,
+        loadSchema: () => Promise.resolve({})
       }
     })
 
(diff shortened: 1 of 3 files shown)


josdejong:
Thanks Stefan, I'll review your PR as soon as I can.

src/lib/plugins/validator/createAjvValidator.test.ts
```diff
@@ -83,9 +83,17 @@ const invalidJson = {
   }
 }
 
-describe('createAjvValidator', () => {
-  test('should create a validate function', () => {
-    const validate = createAjvValidator({ schema, schemaDefinitions })
+function sharedAssertions(factoryFn: typeof createAjvValidator | typeof createAjvValidatorAsync) {
```
josdejong, line 86: Can you replace the solution with a `sharedAssertions` function with using `test.each` of vitest? Docs: https://vitest.dev/api/#test-each

Something like:

```js
describe.each([
  { name: 'createAjvValidator', create: createAjvValidator },
  { name: 'createAjvValidatorAsync', create: createAjvValidatorAsync }
])('$name', ({ create }) => {
  // ... the tests ...
})
```

src/lib/plugins/validator/createAjvValidator.ts
```diff
@@ -1,62 +1,84 @@
...
+ *         $ref: '/schema.json'
+ *       },
+ *       ajvOptions: {
+ *         loadSchema(uri) {
+ *           return fetch(uri).then((res) => res.json())
+ *         }
+ *       }
+ *     })
+ */
+export async function createAjvValidatorAsync(options: AjvValidatorOptions): Promise<Validator> {
```
josdejong, line 70: Can you describe the new function `createAjvValidatorAsync` in the README.md, in the sections "validator" and "Utility functions"?

src/lib/plugins/validator/createAjvValidator.ts
```diff
@@ -1,62 +1,84 @@
...
+   * your own Ajv instance and ignore the provided one.
+   */
   onCreateAjv?: (ajv: Ajv) => Ajv | void
+
+  /**
+   * The severity of the validation error.
+   *
+   * @default ValidationSeverity.warning
+   */
+  errorSeverity?: ValidationSeverity
```
josdejong, line 35: Nice addition 👌

josdejong (changes requested):
Thanks Stefan, your PR looks good! I like the way you've refactored both the code and made sure the tests are not duplicated. It makes sense to create a separate function for the async variant since it returnse a Promise, that would be a breaking change and not something you want to deal with in the synchronous case.

I made a few inline comments, can you have a look at those?

src/lib/plugins/validator/createAjvValidator.test.ts
iStefo: Nice, I didn't know that this is a thing :)

src/lib/plugins/validator/createAjvValidator.ts
iStefo: Good point, I added description an example.

iStefo:
Thanks for your feedback! I made the suggested changes and force-pushed the amended commit to keep the history clean. Please tell me in case I should rebase against main.

src/lib/plugins/validator/createAjvValidator.test.ts
josdejong: Thanks! Now, I think all tests for `createAjvValidator` use an async in front of it, so we do not know for sure whether `createAjvValidator` returns synchronously, right? Can you please add 1 separate test to verify that `createAjvValidator` works synchronous?

josdejong:
> Thanks for your feedback! I made the suggested changes and force-pushed the amended commit to keep the history clean. Please tell me in case I should rebase against main.

Thanks for the updates Stefan! Looks good.

I made two more (minor) inline comments, can you have a look at them?

iStefo:
Valid points, I added your suggestions :)

josdejong (approved):

josdejong:
Thanks! Merging now.

josdejong:
Published in `v3.1.0`

PR #523, opened 2025-02-19
chore: update dependency

mzner:
fixes critical security issue.

https://github.com/advisories/GHSA-pppg-cpfq-h7wr

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,7 +28,7 @@
         "immutable-json-patch": "^6.0.1",
         "jmespath": "^0.16.0",
         "json-source-map": "^0.6.1",
-        "jsonpath-plus": "^9.0.0 || ^10.2.0",
+        "jsonpath-plus": "^10.3.0",
         "jsonrepair": "^3.0.0",
         "lodash-es": "^4.17.21",
         "memoize-one": "^6.0.0",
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks Matteo! Though this security issue isn't a big deal because `svelte-jsoneditor` is just used client side, it's good to update the library. 

I think this issue is also the case in `jsonpath-plus@9.x`, so I think we should drop the `^9.0.0` from package.json too, right?

mzner:
> Thanks Matteo! Though this security issue isn't a big deal because `svelte-jsoneditor` is just used client side, it's good to update the library.
> 
> I think this issue is also the case in `jsonpath-plus@9.x`, so I think we should drop the `^9.0.0` from package.json too, right?

Yea, exactly. Shall I update this package as well?

josdejong:
Yes please, can you change the version of `jsonpath-plus` to just `^10.3.0`? That will be a breaking change.

mzner:
> Yes please, can you change the version of `jsonpath-plus` to just `^10.3.0`? That will be a breaking change.

done.

josdejong:
@mzner can you please remove the `pnpm-lock.yaml` file from the PR?

josdejong:
That was fast 🚀 , thanks!

PR #536, opened 2025-03-21
feat: use `plaintext-only` to prevent rich-text edits when editing a value

josdejong:
Note that rich-edits are not a big deal, since they mostly occur when pasting rich content, but the editor only pastes plain text contents. Still, explictely making the EditableDiv plain text can prevent odd cases.


josdejong:
Thinking about it more, I'll not merge this PR right now: I haven't encountered an actual issue yet, so this doesn't solve something concrete right now, but it _does_ add code and complexity.

PR #538, opened 2025-03-21
Truncate large strings

### LINKED ISSUE #537: [Enhancement] Option to truncate large strings
Hello,

I have some json that sometimes contains dxf files - which means a pretty large string.
This gives me the following issues;

- JsonEditor becomes slow, mostly the initial load
- The large string causes a lot of the information i actually want to see to be removed and instead i just see a wall of text.
Image

I'd love an option to truncate large strings, and show them again when clicking on the item (in tree mode).
For example;

`"MyLargeString": "sdaksmdamdkalsmdaksd<truncated>...`

Or something along those lines.

VsCode handles it this way:

Image
Image

- It would be nice when copying the data i would get the actual json string

josdejong:
Fixes #537

Truncates large strings to render only the first 1000 characters, and shows a "Show more" button to expand the text. When a search result is located in the truncated part of the text, the text will be expanded as soon as this search result becomes the active search result.

To do:

- [x] Truncate large strings
- [x] Expand when the active search item is in the truncated text
- [x] Create a config variable to configure the max length
- [x] Change KB notation into KiB
- [x] Extensive test round, also test performance and search&highlight
- [ ] See if we can improve scrolling to the current active search result in a large string

image


josdejong:
Ok I've addressed the feedbacks. There is one open end: improve scrolling to the active search result. I'll address that in a separate PR (it's not a regression or anything, just something I noticed whilst working on this PR).

Issue #557, opened 2025-07-30
"Collapse All", "Collapse Contents", "Collapse Layer", "Collapse Recursively", and "Collapse Unchanged" Functionality for Tree and Text Mode


Jediweirdo:
Thanks for working on this project! This is hands down the best JSON viewer I've ever used and is perfectly serviceable as it is now. This isn't an "issue" more than it's a feature request. When I collapse an object with multiple layers in Text or Tree mode, it'll preserve the collapsed status of every array/object inside the parent object. While there are times when I find lossless collapsing useful, there are many more when I wish I had more nuanced control over what collapses when. So, I propose these 5 collapse types (which are really just 2 types in a trench coat) for text _and_ tree mode:

**1. Collapse All**
This is already in Tree Mode, but this would be nice to have in text mode as both a button and a keyboard shortcut. 

**2. Collapse Layer**
There are times when I want to edit a dictionary entry I know the location of, but trying to find it quickly gets annoying when there's a gargantuan number of nested arrays and dictionaries between me and my destination. Using the Find/Replace menu helps, but I have to take my hands off the mouse to type the name of the entry. It's a first-world problem, but it gets infinitely more annoying the longer you have to do it. Instead, what if you could alt + click the parent object/array's collapse button to collapse every nested dictionary/array inside of it (but not the parent object itself)? This would basically be equivalent to collapsing every topmost entry by hand.

**3. Collapse Contents**
This is a very similar idea to Collapse layer, but Collapse Contents would recursively collapse the parent entry's contents instead of the normal lossless collapse that saves the collapse state of objects. Instead of alt + click, you would activate it with ctrl + click.

**4. Collapse Recursively**
Another similar type of collapse with a very minor difference. If you hold shift while content or layer collapsing a parent object, it'll collapse the parent as well. Think of it as a self-contained "Collapse all."

**5. Collapse Others**
[...]

PR #558, opened 2025-08-01
feat: support text_mode collapse recursive

CahierX:
To solve https://github.com/josdejong/svelte-jsoneditor/issues/557 issues
Add text_mode collapse recursive

### DIFF
--- a/src/lib/components/__snapshots__/JSONEditor.test.ts.snap
+++ b/src/lib/components/__snapshots__/JSONEditor.test.ts.snap
@@ -626,7 +626,7 @@ exports[`JSONEditor > render text mode 1`] = `
   >
     <!---->
     <div
-      class="jse-text-mode svelte-xt61xw"
+      class="jse-text-mode svelte-1ovs4jk"
     >
       <div
         class="jse-menu svelte-pf7s2l"
@@ -921,8 +921,10 @@ exports[`JSONEditor > render text mode 1`] = `
       </div>
       <!---->
        
+      <!---->
+       
       <div
-        class="jse-contents svelte-xt61xw"
+        class="jse-contents svelte-1ovs4jk"
       >
         <div
           class="cm-editor ͼ1 ͼ2 ͼ4 ͼt"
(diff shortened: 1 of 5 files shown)


josdejong:
That is impressive, this looks promising! 😎 

The code looks really neat. Whilst testing I came across two issues though:

1. When I collapse the whole document with `refJsonEditor.collapse([], true)`, the recursiveness doesn't work: only the root object or array is collapsed.
2. When I open a large document and I perform `refJsonEditor.collapse([], true)`, only the first part of the items in an array is collapsed. For example take [the "10k users" example from here](https://jsoneditoronline.org/indepth/datasets/json-file-example/#a-large-json-document). This may be related to CodeMirror only rendering the visible part of the contents, I'm not sure.

I think issue (1) is likely a small bug but (2) may be tricky.

CahierX:
> That is impressive, this looks promising! 😎
> 
> The code looks really neat. Whilst testing I came across two issues though:
> 
> 1. When I collapse the whole document with `refJsonEditor.collapse([], true)`, the recursiveness doesn't work: only the root object or array is collapsed.
> 2. When I open a large document and I perform `refJsonEditor.collapse([], true)`, only the first part of the items in an array is collapsed. For example take [the "10k users" example from here](https://jsoneditoronline.org/indepth/datasets/json-file-example/#a-large-json-document). This may be related to CodeMirror only rendering the visible part of the contents, I'm not sure.
> 
> I think issue (1) is likely a small bug but (2) may be tricky.

I will try to fix it

CahierX:
I also submitted a [PR](https://github.com/codemirror/language/pull/8) to CodeMirror. If it gets merged, we’ll be able to remove our `createCustomFoldService`.

CahierX:
Sorry for the delay in submitting the code - I've been quite busy with work lately.

Regarding the performance issue with `codeMirrorView.dispatch({effects})` when handling large JSON files, I investigated the CodeMirror source code but couldn't find any viable directions for performance optimization. Therefore, I've implemented solution 2 to address this issue.

I've added a folding progress bar with cancellation support.
Clipboard_Screenshot_1755008872

josdejong:
Thanks, awesome job!

> Are you feeling better now?

Yes 😅. It was just the flu, nothing special though quite annoying.

> As for the points you mentioned (1, 2, and 3), I think they’re all very reasonable. I’ll try to add them in future PRs when I have the time.

👍

CahierX:
> Just for reference: I've added "Expand All" and "Collapse All" buttons to the menu in text mode now, see #563. I love it, thanks again for making this possible @CahierX!

🎉🎉

PR #566, opened 2025-11-03
add setAriaLabel

### LINKED ISSUE #451: How to add associated label to textarea
Hey! i am wondering if there is anyway to add additional attributes to editor and menus. Most importantly, I would like to add a `label` element for the rendered editor (the `div` with the role of `textbox`) for accessibility reasons. I would be able to do so if this element had a unique `id` but i do not see one on the rendered element. Is there a way to add a label that is bound to this textbox, or a way to add additional attributes (like id or even aria attributes) to the rendered elements?

`<div spellcheck="false" autocorrect="off" autocapitalize="off" translate="no" contenteditable="true" style="tab-size: 4;" class="cm-content cm-lineWrapping" role="textbox" aria-multiline="true" data-language="json" aria-autocomplete="list"><div class="cm-activeLine cm-line" style="padding-left: calc(2ch + 6px); text-indent: calc(-2ch + 0px);">{}</div></div>`

steve02081504:
fix #451

### DIFF
--- a/src/lib/components/JSONEditor.svelte
+++ b/src/lib/components/JSONEditor.svelte
@@ -130,6 +130,7 @@
   let jsonEditorModalProps: JSONEditorModalProps | undefined = undefined
   let sortModalProps: SortModalCallback | undefined
   let transformModalProps: TransformModalProps | undefined
+  let ariaLabelMain: string | undefined = undefined
 
   $: {
     const contentError = validateContentType(content)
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks a lot Steve for giving this a try! I've asked for clarification on what is needed exactly regarding `aria-label`. Let's close this PR until we have a better understanding.

PR #578, opened 2026-07-23
fix: add the missing `jse` prefix to three css variables

Stadly:
Three theme variables in `src/lib/themes/defaults.scss` are read without the `jse` prefix that every other variable uses:

| line | variable | read as |
| --- | --- | --- |
| 74 | `$panel-button-color-highlight` | `--panel-button-color-highlight` |
| 101 | `$context-menu-tip-color` | `--context-menu-tip-color` |
| 181 | `$message-success-background` | `--message-success-background` |

`defaults.scss` is what the README points users to for the list of available variables, so `--jse-panel-button-color-highlight`, `--jse-context-menu-tip-color`, and `--jse-message-success-background` look like they should work, but they are silently ignored.

The bundled themes make the same assumption:

- [`src/lib/themes/jse-theme-dark.css:31`](https://github.com/josdejong/svelte-jsoneditor/blob/develop/src/lib/themes/jse-theme-dark.css#L31) sets `--jse-panel-button-color-highlight`
- [`src/routes/themes/jse-theme-big.css:21`](https://github.com/josdejong/svelte-jsoneditor/blob/develop/src/routes/themes/jse-theme-big.css#L21) sets `--jse-message-success-background`

Neither currently has any effect: in the dark theme, highlighted panel buttons (search box, navigation bar, text mode) fall back to `--jse-text-color` instead of the intended `#e5e5e5`.

## The change

Each of the three now reads the prefixed name first and falls back to the unprefixed one, so any existing override using the old name keeps working:

```scss
$panel-button-color-highlight: var(
  --jse-panel-button-color-highlight,
  var(--panel-button-color-highlight, $text-color)
);
```

Happy to drop the deprecated fallbacks and make it a plain rename if you would rather save that for a major release.

### DIFF
--- a/src/lib/themes/defaults.scss
+++ b/src/lib/themes/defaults.scss
@@ -71,7 +71,11 @@ $panel-border: var(--jse-panel-border, $main-border);
 
 $panel-button-color: var(--jse-panel-button-color, inherit);
 $panel-button-background: var(--jse-panel-button-background, transparent);
-$panel-button-color-highlight: var(--panel-button-color-highlight, $text-color);
+// the variable without `jse` prefix is deprecated, it is kept for backward compatibility
+$panel-button-color-highlight: var(
+  --jse-panel-button-color-highlight,
+  var(--panel-button-color-highlight, $text-color)
+);
 $panel-button-background-highlight: var(--jse-panel-button-background-highlight, #e0e0e0);
 
 /* navigation-bar */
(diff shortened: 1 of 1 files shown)


josdejong:
Thanks Magnar for fixing these inconsistencies in CSS variable names! Good idea to keep the old names backward compatible.

PR #580, opened 2026-07-23
feat: add an `ariaLabel` prop to name the editor for screen readers

Stadly:
Picks up #451, and tries to answer the two questions that came up when #566 was closed.

## What is unnamed today

The contents of all three modes have no accessible name, so a screen reader announces the tree/table/text box without saying which editor it belongs to, and consumers cannot query the editor by label:

| mode | element carrying the contents | accessible name |
| --- | --- | --- |
| tree | `div.jse-tree-mode[role="tree"]` | — |
| table | `div.jse-table-mode[role="table"]` | — |
| text | CodeMirror `.cm-content` | — |

## Answering the questions from #566

> wouldn't it be enough to set an `aria-label` with a constant value? Is there need to make it customizable?

A constant is enough for the common case, so the prop *defaults* to `'JSON editor'` — every editor is named without the consumer doing anything. It is customizable because a constant cannot cover two cases: a page with more than one editor (a settings editor and a payload editor are then indistinguishable), and an application that is not in English, where a hard-coded English name is read out in the middle of a localized interface.

> I had the impression that there was need for an `aria-label` on the text editor, not on the JSONEditor as a whole

The label is applied per mode, to the element that actually carries the contents, rather than to the outer container as in #566 — an `aria-label` on the wrapper `div` has no effect, since the element has no role. Switching modes moves the label to the new mode's contents.

## The change

```svelte
<JSONEditor bind:content ariaLabel="Settings" />
```

- `JSONEditor.svelte` — the prop, its default, and a case in the `updateProps` switch so `vanilla-jsoneditor` can set it too
- `JSONEditorRoot.svelte` — forwards it to the three modes
- `TreeMode.svelte` / `TableMode.svelte` — `aria-label` on the `role="tree"` / `role="table"` container
- `TextMode.svelte` — `EditorView.contentAttributes`, wrapped in a `Compartment` and reconfigured on change, following the existing `readOnly`/`tabSize`/`theme` props so the label stays reactive
- `types.ts`, `README.md` — public type and docs

## Verification

`npm run check`, `npm run lint`, and `npm run test` all pass (510 tests). The three DOM snapshots in `JSONEditor.test.ts.snap` pick up the label, one per mode, and a new test asserts a custom value reaches the right element in each mode.

Checked in a browser against `examples/basic_usage_one_way_binding`: the label lands on `[role="tree"]`, `[role="table"]` and `.cm-content` as the mode changes, `getByLabel('JSON editor')` resolves, and changing the prop while text mode is open updates the live CodeMirror instance.

## Not included

Two things I left out to keep this reviewable, happy to add either here or in a follow-up:

- `input.jse-hidden-input` in tree and table mode has no accessible name, which axe reports as a violation (`label`). It is `tabindex="-1"` and only used for paste, so `aria-hidden="true"` would probably be the right fix.
- The nested editor inside `JSONEditorModal` gets no label; the modal has its own heading, so it did not seem necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,14 @@ mode: 'tree' | 'text' | 'table'
 
 Open the editor in `'tree'` mode (default), `'table'` mode, or `'text'` mode (formerly: `code` mode).
 
+#### ariaLabel
+
+```ts
+ariaLabel: string
+```
+
+The accessible name of the editor, applied to the contents of all three modes. Default value is `'JSON editor'`. Give each editor a name of its own when a page contains more than one, so screen reader users can tell them apart.
+
 #### mainMenuBar
 
 ```ts
--- a/src/lib/components/JSONEditor.svelte
+++ b/src/lib/components/JSONEditor.svelte
@@ -61,6 +61,7 @@
   const tabSizeDefault = 4
   const truncateTextSizeDefault = 1000
   const modeDefault = Mode.tree
+  const ariaLabelDefault = 'JSON editor'
   const mainMenuBarDefault = true
   const navigationBarDefault = true
   const statusBarDefault = true
@@ -100,6 +101,7 @@
   export let tabSize: number = tabSizeDefault
   export let truncateTextSize: number = truncateTextSizeDefault
   export let mode: Mode = modeDefault
+  export let ariaLabel: string = ariaLabelDefault
   export let mainMenuBar: boolean = mainMenuBarDefault
   export let navigationBar: boolean = navigationBarDefault
   export let statusBar: boolean = statusBarDefault
@@ -312,6 +314,9 @@
           break
         case 'mode':
           mode = props[name] ?? modeDefault
+          break
+        case 'ariaLabel':
+          ariaLabel = props[name] ?? ariaLabelDefault
           break
         case 'mainMenuBar':
           mainMenuBar = props[name] ?? mainMenuBarDefault
@@ -560,6 +565,7 @@
         {indentation}
         {tabSize}
         {truncateTextSize}
+        {ariaLabel}
         {statusBar}
         {askToFormat}
         {mainMenuBar}
--- a/src/lib/components/JSONEditor.test.ts
+++ b/src/lib/components/JSONEditor.test.ts
@@ -60,6 +60,27 @@ describe('JSONEditor', () => {
     expect(target).toMatchSnapshot()
   })
 
+  test('render a custom aria-label in all modes', () => {
+    const ariaLabel = 'Custom label'
+
+    const labelOf = (mode: Mode, selector: string): string | null | undefined => {
+      const target = document.createElement('div')
+
+      mount(JSONEditor, {
+        target,
+        props: { mode, content, ariaLabel }
+      })
+
+      flushSync() // wait until CodeMirror is rendered during onMount
+
+      return target.querySelector(selector)?.getAttribute('aria-label')
+    }
+
+    expect(labelOf(Mode.tree, '[role="tree"]')).toBe(ariaLabel)
+    expect(labelOf(Mode.table, '[role="table"]')).toBe(ariaLabel)
+    expect(labelOf(Mode.text, '.cm-content')).toBe(ariaLabel)
+  })
+
   test('render table mode', () => {
     const target = document.createElement('div')
 
--- a/src/lib/components/__snapshots__/JSONEditor.test.ts.snap
+++ b/src/lib/components/__snapshots__/JSONEditor.test.ts.snap
@@ -275,6 +275,7 @@ exports[`JSONEditor > render table mode 1`] = `
         class="jse-hidden-input-label svelte-1p86y3c"
       >
         <input
+          aria-hidden="true"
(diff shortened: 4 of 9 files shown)


josdejong:
Thanks @Stadly for picking this up! It makes sense to give the contents of the editor a (configurable) aria label 👍.

Two thoughts: 

1. I think we should try to come up with a solution that is consistent for each of the three modes. Shouldn't the text mode get an aria label attached to the `div.jse-text-mode` element, comparable to the `div.jse-tree-mode` and `div.jse-table-mode`? Is there a specific reason to attach the `aria-label` to `.cm-content`?
2. Can you indeed add an attribute `aria-hidden="true"` to the `input.jse-hidden-input` elements? That makes sense.

Stadly:
Thanks for the review!

**2 (aria-hidden):** done for tree and table (fb365c9). The `input.jse-hidden-input` is an off-screen proxy that only captures paste and holds focus, so it has no content worth announcing — today it surfaces as an unnamed textbox and axe reports a critical `label` violation. `aria-hidden="true"` clears that, and doesn't trip `aria-hidden-focus` since the input is `tabindex="-1"` (out of the tab order).

**1 (text mode placement):** I looked at attaching it to `div.jse-text-mode` for symmetry with tree/table, but I think `.cm-content` is the right target. The reason the label works on `.jse-tree-mode` / `.jse-table-mode` is that those elements carry the semantic role (`tree` / `table`). `div.jse-text-mode` has no role — it's a generic wrapper — and the element that actually receives focus and carries the editing role in text mode is CodeMirror's `.cm-content` (role `textbox`).

Measured in Chrome's accessibility tree, for the `.cm-content` textbox the user edits:

| aria-label placed on | `.cm-content` accessible name |
| --- | --- |
| `.cm-content` (current) | **"Edit as text"** ✓ |
| `.jse-text-mode` (generic div) | **""** — the textbox is left unnamed; the name sits on a generic ancestor a screen reader won't associate with the focused control |

So the consistent principle across all three modes is "name the element that carries the editing role" — `role=tree`, `role=table`, and CodeMirror's `textbox`. That's also what CodeMirror recommends (aria-label via `contentAttributes` on `.cm-content`). Happy to add a short comment in each mode spelling that out if it helps.

josdejong:
(2) Thanks for updating `input.jse-hidden-input` 👍 

(1) I'm OK with keeping the role in text mode attached to `.cm-content`, especially if that's what CodeMirror advises itself too. 

One last thought: right now, in tree and table mode, the role is attached to `div.jse-tree-mode` and `div.jse-table-mode`, which contains both the main menu and either the tree or table. Shouldn't we attach the role and aria-label to `div.jse-contents` instead, so it is attached to the actual element containing the tree/table rather than the container element containing both menu and tree/table?

Stadly:
Good point — the `role="treeitem"` nodes live in `div.jse-contents`, so `role="tree"` on `.jse-tree-mode` also wraps the menu and nav bar today.

Two things make relocating it more than a one-line move: on both modes that element is the interaction root — `role` sits with `tabindex="-1"`, the `keydown`/`mousedown`/`contextmenu` handlers and `refJsonEditor` — so a clean move should carry the focus/keyboard wiring too; and `.jse-contents` is rendered per state (populated / loading), so the role+label need a stable wrapper.

Since that's a refactor of the existing role/focus structure — separate from adding a configurable label — shall we keep this PR scoped to `ariaLabel` on the current role elements, and track the `.jse-contents` move as a follow-up? Happy to open that, or fold it in here if you'd rather, but it'll touch focus/keyboard and the content branches so it deserves its own testing pass.

josdejong:
That is a good point about the keyboard and mouse handlers. I prefer to keep them there though: if a user clicks the empty menu bar, or has just used a menu button and the focus is still on that button, I prefer the keyboard shortcuts (like arrows, paste, undo, etc) to just work without the user having to click inside the editor contents first, because that is very annoying.

So, probably, your approach as it is right now is the best. What do you think? Just keep it like this?

EDIT: alternatively, we could move the handlers to the `.jse-contents`, and just make sure that after clicking a button or clicking the empty menu area, that the focus is put on the hidden input so the quickkeys keep working. Anyway: maybe best to think this through in a separate issue/PR.

Stadly:
Sounds good — let's keep it as is. Agreed that shortcuts working from the menu/buttons matters more than the stricter role nesting, so the handlers staying on `.jse-tree-mode`/`.jse-table-mode` is the right call. Opened #582 to track the `.jse-contents` move (with your note about refocusing the hidden input after menu interactions) so it isn't lost.

josdejong:
Thanks Magnar 👌

Stadly:
Thanks for great review processes, @josdejong! Any chance of a release within the next couple of days? Would be great, so I don't have to ship my own workarounds for the issues. Thanks for a very good editor!

josdejong:
> Thanks for great review processes, @josdejong! Any chance of a release within the next couple of days? Would be great, so I don't have to ship my own workarounds for the issues. Thanks for a very good editor!

I've just published `v3.13.0` with your fixes and improvements, and also updated all dependencies. Thanks for your work!

Issue #238, opened 2023-03-07
Safari Scrolls to Top on Focus (React, Vue)


nick-bigger:
When clicking on the JSON Editor near the end of a long JSON document, the browser force scrolls back to the top of the input requiring scrolling back down. I've managed to reproduce this in the Codesandbox for the React and Vue implementations:

![jsonEditor](https://user-images.githubusercontent.com/22163464/223557700-4501c3a8-5f9a-4048-b3f4-bd38c37464df.gif)

**To Reproduce:**
I have created a fork of the linked React implementation just to add in a long example JSON, but you can reproduce this in the original linked React implementation as well if you'd like (you just need to add in enough JSON to need to scroll).

1. Open https://codesandbox.io/s/svelte-jsoneditor-react-forked-ghipor?file=/src/App.js:301-7438
2. Scroll to bottom of JSON Editor
3. Click on any structure
4. Observe browser force scrolled to the top

**Additional Details**
- After this happens once, it doesn't happen until the page is refreshed. 
- This only seems to occur in Safari on Desktop (I've tested and can't repro in Chrome or Safari for iOS)

<sub><sup>Thank you so much ! We love this library and are hoping for a fix for this as we use it for JSON inputs all over our application.</sup></sub>

josdejong:
Thanks for reporting. The editor has a hidden input field which is used to capture quick keys like arrows to navigate through the contents of the editor. When you click somewhere inside the editor, this input field is given focus. This input field has a class name `jse-hidden-input-label`, and it is styled with `position: fixed, right: 0; top: 0; width: 0; height: 0;`. 

Anyone able to debug why Safari scrolls to top when giving this element focus and find a way to prevent Safari from scrolling up? (Chrome and Firefox do not do that).

nick-bigger:
hi there ! just bumping this as it has been few months with no movement - this causes us daily issues and we haven't found any work around :(

nick-bigger:
Thank you so much ! I can't wait to test 😁 we really appreciate all of your hard work on this awesome package

josdejong:
Fixed now in `v0.17.2`



> Thank you so much ! I can't wait to test 😁 we really appreciate all of your hard work on this awesome package

Thanks! You're welcome. If the project is so important to you, please consider [sponsoring me](https://github.com/sponsors/josdejong/dashboard) to ensure I can keep maintaining and improving the library 😄

PR #581, opened 2026-07-23
fix: do not scroll a scrollable ancestor when the editor gets focus

Stadly:
## The problem

When the editor sits inside a scrollable panel, clicking a node scrolls that panel to the top, away from the node the user just clicked.

Tree and table mode move focus to `input.jse-hidden-input`, which is deliberately kept out of sight ([`TreeMode.scss`](https://github.com/josdejong/svelte-jsoneditor/blob/develop/src/lib/components/modes/treemode/TreeMode.scss#L17-L26)):

```scss
.jse-hidden-input {
  position: fixed;
  top: -10px;
  left: -10px;
  width: 1px;
  height: 1px;
}
```

`position: fixed` normally resolves against the viewport, which puts the input outside every scroll container, and focusing it scrolls nothing. But an ancestor with `transform`, `filter`, `contain: layout`, `will-change: transform`, … becomes the containing block for fixed descendants. The input is then placed at *that ancestor's* top left corner — inside the scrollable panel — and the browser scrolls the panel to reveal it.

This is easy to hit in practice, because dialog/drawer components apply exactly those properties. The case I ran into is [bits-ui](https://github.com/huntabyte/bits-ui), which sets `contain: layout style` inline on its dialog content; the same happens with anything animating a panel via `transform`.

Note this is the same family of problem as #238, fixed then for Safari by giving the input a non-zero size.

## Reproduction

An ordinary `overflow: auto` container is *not* enough — the containing block is what matters:

```svelte
<div style="height: 200px; overflow: auto; contain: layout">
  <p style="height: 100px">spacer</p>
  <JSONEditor content={{ json }} />
</div>
```

Scroll to the bottom, then click a key. Measured on `develop`, with the container scrolled to 132:

| interaction | before | after |
| --- | --- | --- |
| click a key | 132 | **0** |
| double click a value | 132 | **0** |

## The change

`focus({ preventScroll: true })` at the four places that focus the hidden input (two per mode). Revealing that input is never wanted — it is positioned out of sight on purpose — while the editor's own scrolling is untouched.

With the fix, both interactions above leave the container at 132. Keyboard navigation still scrolls the editor's contents to follow the selection: arrowing down 25 rows moves `.jse-contents` from 0 to 368, identical to `develop`.

`npm run check`, `npm run lint` and `npm run test` (509 tests) all pass; no snapshot changed, since the DOM is unaffected.

### DIFF
--- a/src/lib/components/modes/tablemode/TableMode.svelte
+++ b/src/lib/components/modes/tablemode/TableMode.svelte
@@ -708,8 +708,10 @@
     // with just .focus(), sometimes the input doesn't react on onpaste events
     // in Chrome when having a large document open and then doing cut/paste.
     // Calling both .focus() and .select() did solve this issue.
+    // the input is positioned outside of the editor, so revealing it would
+    // scroll a scrollable ancestor away from the contents the user clicked
     if (refHiddenInput) {
-      refHiddenInput.focus()
+      refHiddenInput.focus({ preventScroll: true })
       refHiddenInput.select()
     }
   }
@@ -725,7 +727,7 @@
         selection = removeEditModeFromSelection(selection)
 
         if (hasFocus && refHiddenInput) {
-          refHiddenInput.focus()
+          refHiddenInput.focus({ preventScroll: true })
           refHiddenInput.blur()
         }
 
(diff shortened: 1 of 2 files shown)


josdejong:
Thanks a lot for debugging and fixing this issue, and for your clear explanation! I didn't know about the `preventScroll` option, that is a perfect fix in this case I think.

